Issue #1, opened 2021-04-28
jacrev(jacrev(f)) fails for matmul


Chillee:
```
from functorch import jacrev
N = 5
M = 3
W = torch.randn(N, M)
def f(x):
    return W @ x
inps = (torch.randn(M),)
print(jacrev(jacrev(f))(*inps))
```
```
Traceback (most recent call last):
  File "python_key.py", line 70, in <module>
    print(jacrev(jacrev(f))(*inps))
  File "/opt/anaconda/lib/python3.7/site-packages/functorch-0.0.1a0+e27e16f-py3.7-linux-x86_64.egg/functorch/_src/eager_transforms.py", line 87, in wrapper_fn
    result, = vmap(vjp_fn)(basis)
  File "/opt/anaconda/lib/python3.7/site-packages/functorch-0.0.1a0+e27e16f-py3.7-linux-x86_64.egg/functorch/_src/vmap.py", line 258, in wrapped
    batched_outputs = func(*batched_inputs)
  File "/opt/anaconda/lib/python3.7/site-packages/functorch-0.0.1a0+e27e16f-py3.7-linux-x86_64.egg/functorch/_src/eager_transforms.py", line 72, in wrapper
    retain_graph=retain_graph, create_graph=create_graph)
  File "/home/chilli/fb/pytorch/torch/autograd/__init__.py", line 228, in grad
    inputs, allow_unused, accumulate_grad=False)
RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

zou3519:
Hmm, I guess this should return a Tensor full of zeros. There's nothing wrong with the mechanism, it's just that we need to pass `allow_unused` to autograd and if it returns us "None" as the result (which it will) we need to turn that into a tensor full of zeros.

zou3519:
> Hmm, I guess this should return a Tensor full of zeros. There's nothing wrong with the layering mechanism, it's just that we need to pass `allow_unused` to autograd and if it returns us "None" as the result (which it will) we need to turn that into a tensor full of zeros.

zou3519:
There are two things that are wrong:
1. We need to pass allow_unused to autograd.grad
2. If the outputs are not part of *any* computation graph, then autograd complains even if allow_unused is True. This is the same for the grad transform, even after #5 goes in:

```
import torch
from functorch import grad
x = torch.tensor(1.)
y = torch.tensor(2.)

def unrelated(x):
    return y

# Error!
grad(unrelated)(x)
[...]

PR #6, opened 2021-05-02
adds support for grad if inputs don't depend on output

Chillee:
Fixes https://github.com/zou3519/functorch/issues/5

### DIFF
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -24,7 +24,7 @@ def _create_differentiable(inps, level=None):
             aliased = tensor
             return aliased.requires_grad_()
         raise ValueError(f'Thing passed to transform API must be Tensor,'
-                         f'got {type(tensor_or_tuple_of_tensors)}')
+                         f'got {type(x)}')
     return tree_map(create_differentiable, inps)
 
 def _undo_create_differentiable(inps, level=None):
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -20,9 +20,7 @@ from functorch._C import (
 def _create_differentiable(inps, level=None):
     def create_differentiable(x):
         if isinstance(x, torch.Tensor):
-            tensor = x
-            aliased = tensor
-            return aliased.requires_grad_()
+            return x.requires_grad_()
         raise ValueError(f'Thing passed to transform API must be Tensor,'
                          f'got {type(x)}')
     return tree_map(create_differentiable, inps)
@@ -30,7 +28,6 @@ def _create_differentiable(inps, level=None):
 def _undo_create_differentiable(inps, level=None):
     def unwrap_tensors(x):
         if isinstance(x, torch.Tensor):
-            tensor = x
             return _unwrap_for_grad(x, level)
         assert False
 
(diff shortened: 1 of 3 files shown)


test/test_eager_transforms.py
```diff
@@ -279,6 +279,14 @@ def test_argnums(self, device):
...
         self.assertEqual(gy, x)
 
+    def test_empty_grad(self, device):
+        def f(x):
+            return (x['a']**2.0).sum()
+        inps = ({'a':torch.randn(10) + 3, 'b':torch.randn(10)})
+        grads = grad(f)(inps)
+        self.assertNotEqual(grads['a'].sum(), 0.0)
+        self.assertEqual(grads['b'].sum(), 0.0)
+
```
zou3519, line 289: nit: test_empty_grad -> test_zero_grad. "empty" can refer to an "empty tensor", which is a tensor that does not have storage (because it has a zero-sized dimension somewhere).

zou3519, line 289: Could we add a test case for https://github.com/zou3519/functorch/issues/1 as well? I think it should pass with this change (but if there is something else wrong it does not need to be handled in this PR)

functorch/_src/eager_transforms.py
```diff
@@ -17,28 +17,24 @@
...
-        return tuple(map(partial(_undo_create_differentiable, level=level), tensor_or_tuple_of_tensors))
-    if isinstance(tensor_or_tuple_of_tensors, list):
-        return tuple(map(partial(_undo_create_differentiable, level=level), tensor_or_tuple_of_tensors))
-    assert False
+def _create_differentiable(inps, level=None):
+    def create_differentiable(x):
+        if isinstance(x, torch.Tensor):
+            tensor = x
+            aliased = tensor
+            return aliased.requires_grad_()
```
zou3519, line 25: nit: `return x.requires_grad_()`. I wrote the code this way originally because it used to do a bit more computation rather than just reassign variables

functorch/_src/eager_transforms.py
```diff
@@ -17,28 +17,24 @@
...
+        raise ValueError(f'Thing passed to transform API must be Tensor,'
+                         f'got {type(x)}')
+    return tree_map(create_differentiable, inps)
+
+def _undo_create_differentiable(inps, level=None):
+    def unwrap_tensors(x):
+        if isinstance(x, torch.Tensor):
+            tensor = x
```
zou3519, line 33: This line is unused

functorch/_src/eager_transforms.py
```diff
@@ -128,8 +124,13 @@ def wrapper(*args):
 
             # NB: need create_graph so that backward pass isn't run in no_grad mode
             flat_grad_input = torch.autograd.grad(
-                output, flat_diff_args, create_graph=True)
+                output, flat_diff_args, create_graph=True, allow_unused=True)
 
+            def insert_zeros(grad_arg, orig_arg):
```
zou3519, line 129: nit: "replace_none_with_zeros" is a more descriptive

zou3519 (approved):
lgtm. I think this should fix #1 as well.

To answer your offline question about "[there being a] design decision about whether this is what we want to return" -- JAX and the torch.autograd.functional APIs return zeros when there is no data dependence so I think this behavior is reasonable.

test/test_eager_transforms.py
Chillee: Unfortunately, vjp uses its own call and doesn't use grad.

PR #34, opened 2021-05-19
Add mode dispatch stack

Chillee:
Doesn't currently compose vmap with itself - not totally sure why. Perhaps it's the excludedispatchkeyguard.


functorch/csrc/DynamicLayer.h
```diff
@@ -11,16 +11,18 @@ namespace at {
 namespace functorch {
 
 struct TORCH_API DynamicLayer {
-  DynamicLayer(DispatchKey key, int64_t layerId): key_(key), layerId_(layerId) {}
+  DynamicLayer(DispatchKey key, int64_t layerId, optional<int64_t> batchSize = nullopt): key_(key), layerId_(layerId), batchSize_(batchSize) {}
 
   DispatchKey key() const { return key_; }
   int64_t layerId() const { return layerId_; }
+  int64_t batchSize() const { return *batchSize_; }
```
zou3519, line 18: this should probably `TORCH_INTERNAL_ASSERT(batchSize_.defined())` to avoid future segfaults

functorch/csrc/VmapModeRegistrations.cpp
```diff
@@ -18,6 +22,22 @@ TORCH_LIBRARY_IMPL(_, FuncTorchVmapMode, m) {
   m.fallback(torch::CppFunction::makeFallthrough());
 }
 
+#define TENSOROPTIONSPARAMS c10::optional<c10::ScalarType> dtype, c10::optional<c10::Layout> layout, c10::optional<c10::Device> device, c10::optional<bool> pin_memory
```
zou3519, line 25: lol

functorch/csrc/VmapModeRegistrations.cpp
```diff
@@ -18,6 +22,22 @@ TORCH_LIBRARY_IMPL(_, FuncTorchVmapMode, m) {
...
+
+#define TENSOROPTIONSARGS dtype, layout, device, pin_memory
+
+Tensor randn_mbatching_rule(IntArrayRef shape, TENSOROPTIONSPARAMS) {
+
+    c10::impl::ExcludeDispatchKeyGuard guard(kVmapModeKey);
+    auto maybe_layer = maybeCurrentDynamicLayer();
+    auto shapeVec = shape.vec();
```
zou3519, line 33: nit: avoid dynamic allocations using VmapDimVector or DimVector

functorch/csrc/DynamicLayer.h
```diff
@@ -11,16 +11,18 @@ namespace at {
...
  private:
   DispatchKey key_;
   int64_t layerId_;
+  optional<int64_t> batchSize_;
```
zou3519, line 22: This LGTM to me for now. In the long-term future, the DynamicLayer struct probably shouldn't hold a batchSize_ (the DynamicLayer system shouldn't need to know about implementation details of each transform). Instead we want some way for transform writers to come along and "register" some "per-level metadata".

functorch/csrc/DynamicLayer.cpp
```diff
@@ -254,6 +254,7 @@ void dynamicLayerFrontFallback(const c10::OperatorHandle& op, torch::jit::Stack*
   //   exclude = exclude.remove(DispatchKey::Batched);
   } else if (layer.key() == kBatchedKey) {
     exclude = exclude.remove(kBatchedKey);
+    c10::impl::tls_set_dispatch_key_included(DispatchKey::FuncTorchVmapMode, true);
```
zou3519, line 257: After exiting the dynamic layer system, is FuncTorchVmapMode set? I'm a little worried because I don't see who actually turns FuncTorchVmapMode off. (Although [we do reset the local dispatch keyset in DynamicLayerBack](https://github.com/zou3519/functorch/blob/5d7553b667d55329b38993068f04128ecc8eee33/functorch/csrc/DynamicLayer.cpp#L349-L350), the SaveLocalDispatchKeySet actually restores the state when we're done with the current chain of dispatch.

So for example:
```
vmap(torch.sin)(x)
# Is FuncTorchVmapMode still on here? It shouldn't be but I am worried it is in the current design
torch.randn(3, 3)
```

If this is the case, an easy solution is to have _vmapmode_decrement_nesting always attempt to remove FuncTorchVmapMode from the local include set

functorch/_src/vmap.py
```diff
@@ -86,12 +83,17 @@ def _create_batched_inputs(
                 f'of dimensionality {arg.dim()} so expected in_dim to satisfy '
                 f'0 <= in_dim < {arg.dim()}.')
 
-    batch_size = _validate_and_get_batch_size(flat_in_dims, flat_args)
+    return (_validate_and_get_batch_size(flat_in_dims, flat_args), flat_in_dims, flat_args, args_spec)
```
zou3519, line 86: nit: outer parentheses aren't needed (`return a, b` in python returns a tuple `(a, b)`)

zou3519 (commented):
general structure lgtm, one concern

functorch/csrc/DynamicLayer.h
Chillee: yeah, agreed.

PR #52, opened 2021-06-07
[BC-breaking] Update make_functional*

zou3519:
Updates make_functional to use the new improved variants. The new
variants are superior in every way so we're replacing the previous
variants with this.

If someone wants the older variants, they can be found at:
- make_functional_with_buffers_deprecated_v1
- make_functional_deprecated_v1


zou3519:
Will ship after notifying potential stakeholders.

zou3519:
Warning added based off of Chillee's suggestion

PR #53, opened 2021-06-08
Setup circleci

zou3519:
Scripts taken from pytorch/nestedtensor.


zou3519:
Build failures are caused by a compiler bug: https://godbolt.org/z/sT9GK714s (newer versions of gcc do not have this problem)

Issue #27, opened 2021-05-14
squeeze.dim batching rule doesn't work for scalar tensors


Chillee:
```
def f(x):
    return torch.squeeze(x, dim=0)

vmap(f)(torch.randn((3)))
```

zou3519:
PyTorch has really weird special cases around scalar tensors. Not sure how easy it will be for us to replicate the semantics.

Chillee:
> Not sure how easy it will be for us to replicate the semantics.

I suspect we're going to need to do so regardless.

laurencer:
@zou3519 - what's the expected behavior in this case? I'm trying to figure out what this should actually do.

Reasoning it out:

1. `x = torch.randn((3))` gives a tensor with shape `(3) (dim=1)`.
2. `f(x) : Tensor<{arbitrary shape; dim >= 1}> -> Tensor<{resultant shape; dim >= 1}>`
3. `vmap(f(x)) : Tensor<{arbitrary shape + 1 dim}> -> Tensor<{resultant shape + 1 dim}>`

The types/shapes don't match up for `x`. E.g. `x` only has a single dimension when passed to `vmap(f(x))` and would thus fail? Or should `x` be promoted to `(3, 1)` automatically (or `(1, 3)`) and then passed to `vmap(f(x))`?

This seems like an issue in the general dispatcher (rather than a batching rule registration style). E.g. when you promote a function using `vmap` the resultant function should check for scalar inputs (i.e. `dim=1`) and automatically promote to `dim=2`?

zou3519:
@laurencer good question. For out-of-place operations and ignoring views, vmap(f)(x) should be equivalent to running `torch.stack([f(xi) for xi in x.unbind(0)])`. This heuristic tells us the following:
1. `x = torch.randn((3))` gives a tensor with shape `(3,)` (dim=1)
2. `f(x[0])` gives a tensor with shape `[]` (dim=0), so `torch.stack([f(xi) for xi in x.unbind(0)])` gives a tensor of shape `[3]` (dim=1)
3. so `vmap(f)(x)` should give us a tensor with shape `(3,)` (dim=1)

I think the batching rule for `squeeze` should check to see if the tensor has dim 1 (and that the dim argument is equal to 0). If it does, then it returns an alias of the tensor (via tensor.alias()).

laurencer:
@zou3519 - is there a description of the new versus old style of batching rules?

[...]

PR #81, opened 2021-07-27
Updated squeeze and squeeze.dim batching rules to new style [#27]

laurencer:
Addresses #27 by updating `squeeze` and `squeeze.dim` to the new batching rules style while also adding special casing for scalar arrays to match underlying PyTorch behavior.

### DIFF
--- a/functorch/csrc/BatchingRegistrations.cpp
+++ b/functorch/csrc/BatchingRegistrations.cpp
@@ -448,42 +448,6 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
   return self;
 }
 
-Tensor squeeze_batching_rule(const Tensor& self) {
-  if (!participatesInCurrentLevel(self)) {
-    c10::impl::ExcludeDispatchKeyGuard guard(kBatchedKey);
-    return self.squeeze();
-  }
-  auto self_physical = MultiBatchVmapTransform::logicalToPhysical(self);
-  auto physical_sizes = self_physical.tensor().sizes();
-
-  // Don't squeeze the batch dims!
-  VmapDimVector squeezed_sizes;
-  int64_t num_batch_dims = self_physical.numBatchDims();
-  squeezed_sizes.insert(
-      squeezed_sizes.end(),
-      physical_sizes.begin(),
-      physical_sizes.begin() + num_batch_dims);
-  for (auto it = physical_sizes.begin() + num_batch_dims; it != physical_sizes.end(); ++it) {
-    if (*it != 1) {
-      squeezed_sizes.push_back(*it);
-    }
-  }
-
-  auto result = self_physical.tensor().view(squeezed_sizes);
-  return self_physical.getPhysicalToLogicalMap().apply(result);
-}
-
-Tensor squeeze_dim_batching_rule(const Tensor& self, int64_t dim) {
-  if (!participatesInCurrentLevel(self)) {
-    c10::impl::ExcludeDispatchKeyGuard guard(kBatchedKey);
-    return self.squeeze(dim);
-  }
-  auto self_physical = MultiBatchVmapTransform::logicalToPhysical(self);
-  auto dim_physical = self_physical.getPhysicalDim(dim);
-  auto result = self_physical.tensor().squeeze(dim_physical);
-  return self_physical.getPhysicalToLogicalMap().apply(result);
-}
-
 Tensor trace_batching_rule(const Tensor& self) {
   if (!participatesInCurrentLevel(self)) {
     c10::impl::ExcludeDispatchKeyGuard guard(kBatchedKey);
@@ -1372,8 +1336,6 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   m.impl("slice.Tensor", slice_batching_rule);
   m.impl("split.Tensor", split_batching_rule);
   m.impl("split_with_sizes", split_with_sizes_batching_rule);
-  m.impl("squeeze", squeeze_batching_rule);
-  m.impl("squeeze.dim", squeeze_dim_batching_rule);
   m.impl("squeeze_.dim", squeeze_dim__batching_rule);
   m.impl("t", native::t); // composite wrt autograd
   // m.impl("trace", trace_batching_rule);
--- a/test/test_vmap.py
+++ b/test/test_vmap.py
@@ -2106,16 +2106,36 @@ class TestVmapOperators(Namespace.TestVmapBase):
              (torch.rand(3, 5, B0, B1, B2),), in_dims=2)
 
     def test_squeeze(self):
-        test = self._vmap_view_test
-        op = torch.squeeze
-        B0, B1 = 1, 11
-        test(op, (torch.rand(B0),))
-        test(op, (torch.rand(B0, 3, 5),))
-        test(op, (torch.rand(1, B0, 5),), in_dims=1)
-        test(op, (torch.rand(B0, 0, 1, 5, 1),))
-        test(op, (torch.rand(B0, 1, 1, 1, 1),))
-        test(vmap(op), (torch.rand(B0, B1, 1),))
-        test(vmap(op), (torch.rand(B1, 1, B0),), in_dims=2)
+        def verify_behavior(op, min_ndim=1):
+            test = self._vmap_view_test
+            B0, B1 = 1, 11
+            # These tests cannot be used with an operator that requires more
+            # than 1 dimension after batching.
+            if min_ndim <= 1:
+                test(op, (torch.rand(B0),))
+                test(op, (torch.rand(B1),))
+                test(vmap(op), (torch.rand(B0, B1, 1),))
+                test(vmap(op), (torch.rand(B1, 1, B0),), in_dims=2)
+            test(op, (torch.rand(B0, 3, 5),))
+            test(op, (torch.rand(1, B0, 5),), in_dims=1)
+            test(op, (torch.rand(B0, 0, 1, 5, 1),))
+            test(op, (torch.rand(B0, 1, 1, 1, 1),))
+            test(vmap(op), (torch.rand(B0, B1, 1, 3, 4),))
+            test(vmap(op), (torch.rand(B1, 1, B0, 4, 5),), in_dims=2)
+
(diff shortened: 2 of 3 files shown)


laurencer:
@zou3519 - let me know what you think. I might try to tidy up the tests a bit so there's less copied code but would be good to get your initial review/thoughts on the approach.

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -1372,8 +1415,8 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   m.impl("slice.Tensor", slice_batching_rule);
   m.impl("split.Tensor", split_batching_rule);
   m.impl("split_with_sizes", split_with_sizes_batching_rule);
-  m.impl("squeeze", squeeze_batching_rule);
-  m.impl("squeeze.dim", squeeze_dim_batching_rule);
+  VMAP_SUPPORT("squeeze", squeeze_batching_rule);
+  VMAP_SUPPORT("squeeze.dim", squeeze_dim_batching_rule);
```
zou3519, line 1419: Let's move these functions to `BatchRulesView.cpp` -- squeeze is a view function and BatchingRegistrations.cpp is meant for the old-style batching rules (it's not documented as so though)

test/test_vmap.py
```diff
@@ -2106,16 +2106,41 @@ def test_slice(self):
...
+            test(op, (torch.rand(B0, 0, 1, 5, 1),))
+            test(op, (torch.rand(B0, 1, 1, 1, 1),))
+            test(vmap(op), (torch.rand(B0, B1, 1, 3, 4),))
+            test(vmap(op), (torch.rand(B1, 1, B0, 4, 5),), in_dims=2)
+
+        verify_behavior(torch.squeeze)
+        verify_behavior(lambda x: torch.squeeze(x, dim=1), min_size=2)
+        verify_behavior(lambda x: torch.squeeze(x, dim=-1), min_size=2)
+        verify_behavior(lambda x: torch.squeeze(x, dim=-2), min_size=3)
+
```
zou3519, line 2130: I think we can uncomment out https://github.com/facebookresearch/functorch/blob/b73213bad5c5c6aa5eeaa96a048bf1163f4cdb40/test/test_vmap.py#L2908. PyTorch has an operator database and we have some autogenerated tests for each operator in that database. Our autogenerated tests for vmap aren't as comprehensive as manually writing tests yet (so the tests you wrote are still helpful!)

test/test_vmap.py
```diff
@@ -2106,16 +2106,41 @@ def test_slice(self):
...
-        B0, B1 = 1, 11
-        test(op, (torch.rand(B0),))
-        test(op, (torch.rand(B0, 3, 5),))
-        test(op, (torch.rand(1, B0, 5),), in_dims=1)
-        test(op, (torch.rand(B0, 0, 1, 5, 1),))
-        test(op, (torch.rand(B0, 1, 1, 1, 1),))
-        test(vmap(op), (torch.rand(B0, B1, 1),))
-        test(vmap(op), (torch.rand(B1, 1, B0),), in_dims=2)
+        assert vmap(op)(x).equal(torch.stack([op(xi) for xi in x.unbind(0)]))
+        test(op, (x,))
```
zou3519, line 2138: `_vmap_view_test` should already check `assert vmap(op)(x).equal(torch.stack([op(xi) for xi in x.unbind(0)]))`

test/test_vmap.py
```diff
@@ -2106,16 +2106,41 @@ def test_slice(self):
              (torch.rand(3, 5, B0, B1, B2),), in_dims=2)
 
     def test_squeeze(self):
+        def verify_behavior(op, min_size=1):
```
zou3519, line 2109: By size do you mean "number of dimensions"? If so we should name this `min_rank` or `min_ndim`.

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -448,40 +448,83 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
...
-  int64_t num_batch_dims = self_physical.numBatchDims();
-  squeezed_sizes.insert(
-      squeezed_sizes.end(),
-      physical_sizes.begin(),
-      physical_sizes.begin() + num_batch_dims);
-  for (auto it = physical_sizes.begin() + num_batch_dims; it != physical_sizes.end(); ++it) {
-    if (*it != 1) {
-      squeezed_sizes.push_back(*it);
+std::tuple<Tensor, optional<int64_t>> squeeze_batching_rule(const Tensor& self, optional<int64_t> bdim) {
+  if (bdim) {
```
zou3519, line 452: I forgot to mention this, but `bdim` is never nullopt in an operator that accepts a single Tensor. If bdim is nullopt, then the batching rule doesn't actually get invoked; PyTorch dispatcher ends up not dispatching to the Batched implementation of things. Because of this we can add `TORCH_INTERNAL_ASSERT(bdim.has_value())` and delete the branch where bdim is nullopt.

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -448,40 +448,83 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
...
+    if (self.dim() == 1) {
+      // Special case for scalar arrays to replicate PyTorch behavior.
+      return std::make_tuple<Tensor, optional<int64_t>>(
```
zou3519, line 455: nit: I think `make_tuple` can deduce the type of its template so no need to write it out: `std::make_tuple(...)`

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -448,40 +448,83 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
...
+          // Only increment for the dimensions that will be kept in the output.
+          if (before_batch_idx) {
+            ++new_batch_idx;
+          }
+        }
+        ++original_idx;
+      }
+
+      auto result = self.view(squeezed_sizes);
+      return std::make_tuple<Tensor, optional<int64_t>>(std::move(result), c10::optional<int64_t>(new_batch_idx));
```
zou3519, line 486: nit: I think `make_tuple` can deduce the type of its template so no need to write it out: `std::make_tuple(...)`

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -448,40 +448,83 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
...
+  } else {
+    // The array does not have a batch dimension so we can return using the underlying squeeze function.
+    return std::make_tuple<Tensor, optional<int64_t>>(self.squeeze(), c10::optional<int64_t>());
+  }
+}
+
+std::tuple<Tensor, optional<int64_t>> squeeze_dim_batching_rule(const Tensor& self, optional<int64_t> bdim, int64_t dim) {
+  if (bdim) {
+    if (self.dim() == 1) {
+      // Special case for scalar arrays to replicate PyTorch behavior.
```
zou3519, line 497: We should TORCH_CHECK that `dim == 0` here. PyTorch throws an error when that is not the case while squeezing a scalar:
```
x = torch.tensor(1.)
x.squeeze(1)
```

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -448,40 +448,83 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
...
+      return std::make_tuple<Tensor, optional<int64_t>>(
+        std::move(self.alias()),
+        std::move(bdim)
+      );
+    } else {
+      // Calculate the proper offset if dim is negative.
+      auto actual_dim = dim;
+      if (dim < 0) {
+        actual_dim = self.dim() + dim - 1;
+      }
```
zou3519, line 507: We can use maybe_wrap_dim here: https://github.com/pytorch/pytorch/blob/7c588d5d00177230ac98bd92510e39bd0ce6f2da/aten/src/ATen/WrapDimUtils.h#L9-L11

I think the usage is `actual_dim = maybe_wrap_dim(dim, self.dim())` but there might be a off-by-one error in there

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -448,40 +448,83 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
...
-      squeezed_sizes.push_back(*it);
+std::tuple<Tensor, optional<int64_t>> squeeze_batching_rule(const Tensor& self, optional<int64_t> bdim) {
+  if (bdim) {
+    if (self.dim() == 1) {
+      // Special case for scalar arrays to replicate PyTorch behavior.
+      return std::make_tuple<Tensor, optional<int64_t>>(
+        std::move(self.alias()),
+        std::move(bdim)
+      );
+    } else {
```
zou3519, line 459: nit: this is a style thing, but we can reduce the levels of indentation:
```
if (self.dim() == 1) {
   return ...;
}
// handle the other case here
```

functorch/csrc/BatchingRegistrations.cpp
```diff
@@ -448,40 +448,83 @@ Tensor& zero_inplace_batching_rule(Tensor &self) {
...
-  for (auto it = physical_sizes.begin() + num_batch_dims; it != physical_sizes.end(); ++it) {
-    if (*it != 1) {
-      squeezed_sizes.push_back(*it);
+std::tuple<Tensor, optional<int64_t>> squeeze_batching_rule(const Tensor& self, optional<int64_t> bdim) {
+  if (bdim) {
+    if (self.dim() == 1) {
+      // Special case for scalar arrays to replicate PyTorch behavior.
+      return std::make_tuple<Tensor, optional<int64_t>>(
+        std::move(self.alias()),
+        std::move(bdim)
```
zou3519, line 457: Do we need std::move for these? (I'm not very good with C++ so I don't know and the rest of this repo has not been using std::move this way)

zou3519 (commented):
This looks pretty good! I had some minor comments

laurencer:
@zou3519 - thanks for the review. I pushed the changes that address your feedback!

zou3519 (approved):
LGTM, thank you @laurencer!

Issue #85, opened 2021-08-02
Can't implement batching rule for F.one_hot as written


Chillee:
We can't currently implement a batching rule for `F.one_hot`, even if we had batching rules for the constituent ops (`_local_scalar_dense`, `min`, `max`, `unsqueeze`, `scatter`).

In the C++ code, the code looks like this here (https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/Onehot.cpp#L38)
```
    Tensor ret = at::zeros(shape, self.options());
    ret.scatter_(-1, self.unsqueeze(-1), 1);
    return ret;
```
cc: @bdhirsh

pedro-abreu:
Any ideas how to solve this? I've been using vmap and found this severely impacted my performance.

Chillee:
@pedro-abreu We can prioritize fixing it - there's a couple different ways we can resolve it.

One of them is by decomposing it into a `zeros` + non-inplace scatter. We can do that.

zou3519:
@pedro-abreu we've added a quick fix in the form of https://github.com/facebookresearch/functorch/pull/93. If you install functorch again using the latest main branch that should do the job. Please let us know if there are other batching rules we can help implement for your use case!

pedro-abreu:
Thank you, I will test it soon, so you can close this issue.

pedro-abreu:
My performance issues are fixed and batching works fine now :) Thank you!

zou3519:
@pedro-abreu out of curiosity (and to collect some feedback), could you please tell us a little bit about what you're using functorch for?

pedro-abreu:
@pedro-abreu trying to implement a differentiable top-k :)

zou3519:
@pedro-abreu that sounds cool! Does that mean that [torch.topk](https://pytorch.org/docs/stable/generated/torch.topk.html) was not sufficient for your use case?

pedro-abreu:
@pedro-abreu Yes, basically you cannot backpropagate through torch.topk. But I think from my ignorant perspective of the underlying codebase, is it possible to batch a custom autograd function with functorch? In JAX you can do a custom vjp by providing defvjp() with a forward and backward function and easily vmap the function where you called defvjp().

zou3519:
> is it possible to batch a custom autograd function with functorch? In JAX you can do a custom vjp by providing defvjp() with a forward and backward function and easily vmap the function where you called defvjp() (not sure if the vjp in functorch allows for this).

[...]

PR #93, opened 2021-08-17
Adds one_hot decomposition, scatter batch rule

zou3519:
The one hot decomposition is a quick hack for #85. I think we might
want to make one_hot a primitive w.r.t. autograd because of the
data-dependent error checking.

Test Plan:
- wait for CI


zou3519:
idk where circleci is, so yolo merge time. The relevant tests passed locally.

PR #123, opened 2021-09-15
Added Index.Tensor batching rule

Chillee:
I also needed to augment the OpInfo tests to reach into pytrees.

There's still a couple of `__getitem__` xfails, but that's because of the dynamic shape boolean masking in some cases :/


zou3519:
It looks like there may have been a rebase problem

PR #127, opened 2021-09-18
Support functions with multiple outputs in `compiled_function`

bertmaher:
Summary: The existing code assumed a single output; this generalizes to tuple
outputs

Test Plan: Compile a simple test program with multiple outputs and check that
outputs/grads are the same as eager.
```
def foo(a, b):
  return a + b, a * b
```

### DIFF
--- a/functorch/_src/eager_compilation.py
+++ b/functorch/_src/eager_compilation.py
@@ -119,6 +119,7 @@ def compiled_function(fn, fw_compiler, bw_compiler):
     compiled_fw = None
     bw_module = None
     compiled_bw = None
+    num_outs = None
 
     saved_fn = None
     def returned_function(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


functorch/_src/eager_compilation.py
```diff
@@ -134,14 +135,26 @@ def vjpfull(primals, tangents):
                 out, vjpfn = vjp(flat_fn, *primals)
                 return out, vjpfn(*tangents)
 
+            def outputs_like(outs):
```
Chillee, line 138: I think we should probably use pytrees here? I suppose it's possible that we actually don't need to do anything at all... the only reason this is here is to end up initializing the trace with tensors the same shape as the outputs.

Chillee: So... we could just pass the outputs in directly to the trace.

Chillee (approved):
Otherwise, looks good to me :)

Probably add a test in `test_pythonkey.py` - don't worry about the OpInfo tests, a manual test will do fine here imo.

The OpInfo tests are probably what we want to use long term, but there is some ... jank there.

PR #128, opened 2021-09-20
Grid sample batch rule registration

samdow:
Adds batch rule for [grid sample](https://pytorch.org/docs/master/generated/torch.nn.functional.grid_sample.html#torch.nn.functional.grid_sample)

### DIFF
--- a/functorch/csrc/BatchRulesModules.cpp
+++ b/functorch/csrc/BatchRulesModules.cpp
@@ -202,28 +202,24 @@ std::tuple<Tensor,Tensor> cudnn_convolution_backward_plumbing(const Tensor & sel
 
 std::tuple<Tensor,optional<int64_t>>
 grid_sample_batch_rule(const Tensor& input, optional<int64_t> input_bdim, const Tensor& grid, optional<int64_t> grid_bdim, int64_t interpolation_mode, int64_t padding_mode, bool align_corners) {
-  std::vector<int64_t> input_spec {0, 1, 2, 3};
-  std::vector<int64_t> grid_spec = input_spec;
-  std::vector<int64_t> out_spec = input_spec;
-
   std::tuple<Tensor, optional<int64_t>> result;
   if (input_bdim && !grid_bdim) {
-    auto new_input = reshape_dim_into(*input_bdim, input_spec[1], input);
+    auto new_input = reshape_dim_into(*input_bdim, 1, input);
     auto out = at::grid_sampler(new_input, grid, interpolation_mode, padding_mode, align_corners);
-    out = reshape_dim_outof(out_spec[1], input.sizes()[*input_bdim], out);
-    result = std::make_tuple(out, out_spec[1]);
+    out = reshape_dim_outof(1, input.sizes()[*input_bdim], out);
+    result = std::make_tuple(out, 1);
   } else if (!input_bdim && grid_bdim) {
     // grid of N(BH)W2 -> NC(BH)W or grid of N(BD)HBW3 -> NC(BD)HW
-    auto new_grid = reshape_dim_into(*grid_bdim, grid_spec[1], grid);
+    auto new_grid = reshape_dim_into(*grid_bdim, 1, grid);
     auto out = at::grid_sampler(input, new_grid, interpolation_mode, padding_mode, align_corners);
-    out = reshape_dim_outof(out_spec[2], grid.sizes()[*grid_bdim], out);
-    result = std::make_tuple(out, out_spec[2]);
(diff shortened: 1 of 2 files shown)


test/test_vmap.py
```diff
@@ -3008,7 +3036,6 @@ def test_vmap_exhaustive(self, device, dtype, op):
         xfail('nn.functional.cross_entropy', 'mean'),
         xfail('nn.functional.cross_entropy', 'none'),
         xfail('nn.functional.cross_entropy', 'sum'),
-        xfail('nn.functional.grid_sample'),
```
samdow, line 3011: I took this skip out but left in the skip in [text_ops.py](https://github.com/facebookresearch/functorch/blob/grid_sample_registration/test/test_ops.py#L444) because that test skips linear and conv2d later down. Not sure if there's something else I need to do to make that test work, but wanted to call it out

functorch/csrc/BatchRulesModules.cpp
```diff
@@ -200,6 +200,66 @@ std::tuple<Tensor,Tensor> cudnn_convolution_backward_plumbing(const Tensor & sel
...
+    out = reshape_dim_outof(out_spec[0], input.sizes()[*grid_bdim], out);
+    result = std::make_tuple(out, out_spec[0]);
+  } else {
+    result = std::make_tuple(at::grid_sampler(input, grid, interpolation_mode, padding_mode, align_corners), nullopt);
+  }
+  return result;
+}
+
+std::tuple<Tensor,optional<int64_t>>
+cudnn_grid_sample_batch_rule(const Tensor& input, optional<int64_t> input_bdim, const Tensor& grid, optional<int64_t> grid_bdim) {
```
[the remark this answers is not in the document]
Chillee: I actually wonder if this is a more ... generic pattern. Hmmm....

Would it be possible to add a top level comment explaining the general rationale behind this pattern?

test/test_vmap.py
Chillee: That test skips linear and conv2d probably because they break :P Specifically for conv, it probably looks for a batching rule for conv_backward as well, which I don't think we have right now.

test/test_vmap.py
```diff
@@ -2548,6 +2548,34 @@ def test_conv2d(self):
             for loop_out, batched_out in get_fallback_and_vmap_exhaustive(conv_fn, arg_values, kwarg_values):
                 self.assertEqual(loop_out, batched_out)
 
+    def test_grid_sample2d(self):
+        N, C, DIn, HIn, WIn, DOut, HOut, WOut = range(1, 9)
```
Chillee, line 2552: More tests are always good, but are these testing anything that isn't covered by the OpInfo tests?

Chillee (approved):
Looks mostly good to me, some minor comments.

test/test_vmap.py
samdow: This is based on the one above for conv2d right above it (we could probably condense these with OpInfo?) . As I understood (which might be wrong), the OpInfo tests that already exist just test that the function didn't call the fallback and this tests correctness of the output (comparing a looped version to the vmap version)

Chillee: There are a bunch of OpInfo tests :P

We have some OpInfo tests that just test whether it has a batching rule (`test_op_has_batch_rule`), and some OpInfo tests that test actual correctness (`test_vmap_exhaustive`).

I think those conv2d tests were added when we didn't have a conv2d OpInfo.

functorch/csrc/BatchRulesModules.cpp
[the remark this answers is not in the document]
samdow: Added [a comment](https://github.com/facebookresearch/functorch/pull/128/files#diff-e5658af94b6758a9726b8fade288818cb136f7dca498110a34973f2481b94c67R204-230) explaining the shapes in the 3 cases (for both 2d and 3d). Let me know if there's more needed to clarify it since it can be hard to read since both the input and grid have Hs and Bs

I was able to make it generic by basically copying what's done for binary op helper (the difference between the convolutions is that the non-cudnn version has more arguments about like padding and stuff). I'm not sure if we can make it more generic with conv (if that's what you were thinking or suggesting) since conv uses group convolutions to make the shapes work at least once and has to deal with bias while grid sampler never has a bias.

samdow:
Looking at failure but JFYI that it looks like it's introduced because of the new nightly version of pytorch (I see it locally when running on main). Don't totally want to just disable the whole conv2d tests so trying to get a better idea of what's up

Chillee:
@samdow yeah feel free to ignore it for landing this PR.

PR #130, opened 2021-09-22
Ported `_reshape_alias` to the new API

vfdev-5:
Description:
- Tried to port `_reshape_alias` to the new API. Algorithm is not exactly the same as in the old method, maybe unsafe ?

cc @zou3519

### DIFF
--- a/functorch/csrc/BatchRulesViews.cpp
+++ b/functorch/csrc/BatchRulesViews.cpp
@@ -274,13 +274,13 @@ std::tuple<Tensor, optional<int64_t>> select_batching_rule(const Tensor& self, o
 
 std::tuple<Tensor, optional<int64_t>> _reshape_alias_batch_rule(const Tensor& self, optional<int64_t> bdim, const IntArrayRef shape, const IntArrayRef strides) {
   (void) strides;
-  if (!bdim) {
-    return std::make_tuple(self.reshape(shape), nullopt);
-  }
+  TORCH_INTERNAL_ASSERT(bdim.has_value());
 
-  auto new_shape = shape.vec();
-  new_shape.insert(new_shape.cbegin() + bdim.value(), self.size(bdim.value()));
-  return std::make_tuple(self.reshape(new_shape), bdim);
+  c10::SmallBuffer<int64_t, 5> new_shape(shape.size() + 1);
+  new_shape[*bdim] = self.size(*bdim);
+  std::copy(shape.begin(), shape.begin() + *bdim, new_shape.begin());
+  std::copy(shape.begin() + *bdim, shape.end(), new_shape.begin() + *bdim + 1);
+  return std::make_tuple(at::reshape(self, new_shape), bdim);
 }
 
 std::tuple<Tensor,optional<int64_t>> diagonal_backward_batch_rule(
--- a/functorch/csrc/BatchRulesViews.cpp
+++ b/functorch/csrc/BatchRulesViews.cpp
@@ -272,6 +272,17 @@ std::tuple<Tensor, optional<int64_t>> select_batching_rule(const Tensor& self, o
   return std::make_tuple(result, 0);
 }
 
+std::tuple<Tensor, optional<int64_t>> _reshape_alias_batch_rule(const Tensor& self, optional<int64_t> bdim, const IntArrayRef shape, const IntArrayRef strides) {
+  (void) strides;
+  if (!bdim) {
+    return std::make_tuple(self.reshape(shape), nullopt);
+  }
+
+  auto new_shape = shape.vec();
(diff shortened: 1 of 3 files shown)


functorch/csrc/BatchRulesViews.cpp
```diff
@@ -272,6 +272,17 @@ std::tuple<Tensor, optional<int64_t>> select_batching_rule(const Tensor& self, o
   return std::make_tuple(result, 0);
 }
 
+std::tuple<Tensor, optional<int64_t>> _reshape_alias_batch_rule(const Tensor& self, optional<int64_t> bdim, const IntArrayRef shape, const IntArrayRef strides) {
+  (void) strides;
+  if (!bdim) {
+    return std::make_tuple(self.reshape(shape), nullopt);
+  }
```
zou3519, line 279: The way we do dispatches means that we don't actually need this case: if none of the inputs have bdims, then the batching rule doesn't get called. If you want to assert this you could add a TORCH_INTERNAL_ASSERT(bdim.has_value()).

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -272,6 +272,17 @@ std::tuple<Tensor, optional<int64_t>> select_batching_rule(const Tensor& self, o
...
+
+  auto new_shape = shape.vec();
+  new_shape.insert(new_shape.cbegin() + bdim.value(), self.size(bdim.value()));
+  return std::make_tuple(self.reshape(new_shape), bdim);
+}
```
zou3519, line 284: This algorithm looks correct to me. Did you have concerns about it, @vfdev-5?

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -272,6 +272,17 @@ std::tuple<Tensor, optional<int64_t>> select_batching_rule(const Tensor& self, o
   return std::make_tuple(result, 0);
 }
 
+std::tuple<Tensor, optional<int64_t>> _reshape_alias_batch_rule(const Tensor& self, optional<int64_t> bdim, const IntArrayRef shape, const IntArrayRef strides) {
+  (void) strides;
+  if (!bdim) {
+    return std::make_tuple(self.reshape(shape), nullopt);
+  }
+
+  auto new_shape = shape.vec();
```
zou3519, line 281: nit: new_shape is a std::vector and std::vector uses dynamic allocations. We should prefer to use SmallBuffer or SmallVector whenever possible.

Example usage: https://github.com/facebookresearch/functorch/blob/43889f5e8ec22f1b5de914232aa9284911adb1fb/functorch/csrc/BatchRulesViews.cpp#L282

Concretely, we want to preallocate a SmallBuffer<int64_t, 5> with (shape.size() + 1) and then std::copy into it and finally use that as the argument to reshape

zou3519 (commented):
This looks pretty good to me. I had some comments and nits

functorch/csrc/BatchRulesViews.cpp
vfdev-5: Well, I was not sure if this covers all cases. I also saw `_unsafe_view_batch_rule` which does almost the same thing we would like with this `_reshape_alias`:
https://github.com/facebookresearch/functorch/blob/43889f5e8ec22f1b5de914232aa9284911adb1fb/functorch/csrc/BatchRulesViews.cpp#L133-L141

zou3519: I think this should cover all the cases, we have tests for that :) (try to run `pytest test/ -v -k "reshape"`). 

Re: unsafe_view: `_unsafe_view` is actually the name for the operator. `_unsafe_view_batch_rule` is not saying that the batching rule is unsafe, it is saying that it is the batching rule for the "`unsafe_view`" operator

vfdev-5:
@zou3519 I updated the code and also checked locally the test. Seems like Circle CI is not configured to run on forks as no tests are running ?

Chillee:
@vfdev-5 Could you try running all pytests?

zou3519:
Hmm, let me try to close and reopen this PR and see if CircleCI picks it up

Chillee:
Seems like it works modulo broken master tests?

zou3519:
Yup, agree. Will merge, thanks for the contribution, @vfdev-5 !

zou3519 (approved):

vfdev-5:
@zou3519 Thanks for the guidance and the review !

zou3519:
We're happy to review more PRs if you're interesting in contributing more :). Horace wrote up a batching rule contribution guide at https://github.com/facebookresearch/functorch/blob/main/writing_batching_rules.md#cool-im-convinced-and-i-want-to-write-batching-rules-where-do-i-find-some that you've probably seen already

PR #133, opened 2021-09-23
Introduce compiled_module for eager compilation of modules

bertmaher:


bertmaher:
I'm a little shocked this works tbh :-).  I had mixed success using tvm_compile here, for what it's worth.  It works for the "fuse_module" example here, but when I tried something fairly simple like (Linear(), Relu()), TVM blows up because `aten::threshold_backward` isn't implemented in the Relay adapter.

Chillee:
> I'm a little shocked this works tbh :-).

Haha, that's one of the nice things about tracing :) Some things become shockingly easy.

> when I tried something fairly simple like (Linear(), Relu()), TVM blows up because aten::threshold_backward isn't implemented in the Relay adapter.

Unfortunately,  that's one of the things we need to do - get compiler backends to implement these backwards ops or provide decompositions of these backwards ops into other existing operators.

functorch/_src/eager_compilation.py
```diff
@@ -182,3 +182,16 @@ def backward(ctx, *args):
...
+    fmod, fparams = make_functional(mod)
+    compiled_f = compiled_function(fmod, fw_compiler, bw_compiler)
+
+    class CompiledModule(nn.Module):
+        def __init__(self):
+            super(CompiledModule, self).__init__()
+            self.orig_module = mod
+
+        def forward(self, *args, **kwargs):
+            return compiled_f(fparams, *args, **kwargs)
```
Chillee, line 196: I'm a little bit worried that this isn't going to work properly.

I suspect we probably need to pull off the params on every invocation.

bertmaher: Hm, really?  What do you think would go wrong versus this approach?  And does "pulling off the params" mean we would need to do a bunch of the work in `make_functional` every time we call a compiled module?

Chillee: Imagine

```
module = compiled_module(module)
module(args)
module.weight = <new param>
module(args)
```

I think this approach would continue to hold a reference to the old param of `module.weight`.

bertmaher: Okay yeah this approach is definitely broken, and I wasn't testing it thoroughly enough.  I think I don't understand the reference vs value semantics at work here; I think the "pulling off of params" done by make_functional must be giving me a new value, because when I update the params (e.g., `mod.param -= mod.param.grad`) it doesn't get reflected in `fparams`.

I guess I have to re-extract the params on every forward, like you suggested?

PR #136, opened 2021-09-23
Added support for negative in_dims

### LINKED ISSUE #134: vmap with negative in_dims ?
I'm wondering if it would make sense to specify and cover `in_dims` as negative value in `vmap` ? For example:
```python
from functorch import vmap

def f(x):
    return torch.sin(x).sum()

t_in = torch.randn(4, 4)
t_out = vmap(f, in_dims=-1)(t_in)
```
Currently, this is not supported
```
ValueError: vmap(f, in_dims=-1, ...)(<inputs>): Got in_dim=-1 for some input, but that input is a Tensor of dimensionality 2 so expected in_dim to satisfy 0 <= in_dim < 2.
```

vfdev-5:
Description:
- Added support for negative in_dims
- Added a basic test

Fixes #134

### DIFF
--- a/functorch/_src/vmap.py
+++ b/functorch/_src/vmap.py
@@ -70,7 +70,7 @@ def _process_batched_inputs(in_dims: in_dims_t, args: Tuple, func: Callable) -> 
             f'in_dims has structure {tree_flatten(in_dims)[1]} but inputs '
             f'has structure {args_spec}.')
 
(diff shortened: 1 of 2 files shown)


Chillee:
First impression is that it looks good :)

I'd approve but a bit sleepy - I'll take another look tomorrow.

Chillee (approved):
LGTM!

PR #137, opened 2021-09-23
Ported `view` op to the new API

vfdev-5:
Description:
- Ported `view` op to the new API
- Updated test_view

Note: code to create new size/shape is the same as in _reshape_alias. Maybe we can refactor that into a separate function ?


Chillee:
> Maybe we can refactor that into a separate function ?

Yeah, I think that's a good idea. Can you do that for this PR?

vfdev-5:
Tests have run here: https://app.circleci.com/pipelines/github/facebookresearch/functorch?branch=pull%2F137
But there is a segfault: https://app.circleci.com/pipelines/github/facebookresearch/functorch/261/workflows/cb09c187-a92b-4cba-aeed-6c209d34f2c6/jobs/1543?invite=true#step-106-133

test/test_vmap.py
```diff
@@ -2438,10 +2434,6 @@ def test_view_as(self):
         B0, B1, B2 = 7, 11, 13
         op = torch.Tensor.view_as
 
-        # We should error out if the view would produce an incorrect result
```
Chillee, line 2441: Why did you remove these tests?

Chillee:
The test errors seem kinda real? Since they're segfaulting on every test.

test/test_vmap.py
vfdev-5: I'm not sure, but if this check is present I have locally:
```
        # We should error out if the view would produce an incorrect result                                                                                                                         with self.assertRaises(RuntimeError):                                                                                                                                               
>           vmap(op, in_dims=(1, 0))(torch.rand(2, B0, 5), torch.rand(B0, 10))                                                                                                              
E           AssertionError: RuntimeError not raised
```

vfdev-5:
@Chillee can you please try close/reopen to schedule tests. Seems like this trick does not work for me

Chillee:
Looks like the CI passes. LGTM!

PR #138, opened 2021-09-23
Added rot90 batch rule

vfdev-5:
Description:
- Added rot90 batch rule
- Enabled associated tests

Related to https://github.com/facebookresearch/functorch/issues/112

Note: tests seem not be reliable to output bdim.
Tests are passing even if batching rule outputs wrong bdim


zou3519:
> Note: tests seem not be reliable to output bdim.
> Tests are passing even if batching rule outputs wrong bdim

Yeah our vmap tests don't test for non-zero bdim. We should fix that at some point...

Chillee:
Does this work with `REDUCTION_BOXED`?

Specifically, `REDUCTION_BOXED_ARGS(rot90, 2)` (which specifies that the dim argument is at position 2).

vfdev-5:
@Chillee I like this magical macro ! Yes, seems like tests are passing with REDUCTION_BOXED_ARGS and without writing any batching rule :)

Chillee:
There's an xfail not passing - I suspect you need to also remove the test from the `has_batch_rule` OpInfo test.

Chillee:
LGTM. Thanks!

PR #141, opened 2021-09-24
Use pytest for testing in CI

zou3519:
Also try out the parallel option with `-n2`.


zou3519:
Conclusion:
- CUDA memory leak testing is not parellel-safe
- should just use regular pytest

Issue #101, opened 2021-08-24
Most used `torch.<x>` operations: 100-200


Chillee:
- [ ] fft	4436
- [x] lt	4421
- [x] log2	4316
- [x] enable_grad	4238
- [x] rand_like	4187
- [ ] argsort	3972
- [x] seed	3932
- [x] mv	3547
- [ ] ger	3309
- [x] ge	3248
- [x] atan2	3210
- [x] ceil	3202
- [x] ne	3075
- [ ] bincount	3063
- [x] acos	3055
- [x] rsqrt	3031
- [ ] svd	3029
- [x] numel	3003
- [x] log1p	2840
- [x] unbind	2808
- [x] le	2714
- [ ] isinf	2707
- [ ] cross	2646
- [x] set_default_dtype	2536
- [x] argmin	2535
- [ ] sparse_coo_tensor	2489
- [x] log10	2304
- [ ] kthvalue	2192
- [x] set_rng_state	2158
- [x] get_rng_state	1996
- [x] get_default_dtype	1879
- [ ] det	1868
- [ ] qr	1864
- [ ] histc	1852
- [ ] symeig	1832
- [x] trace	1801
- [x] median	1795
- [x] addcmul	1751
- [x] remainder	1717
- [ ] baddbmm	1693
- [x] lgamma	1665
- [ ] repeat_interleave	1598
- [x] fmod	1576
- [x] reciprocal	1575
- [x] tan	1560
- [x] initial_seed	1532
- [ ] take	1529
- [ ] stft	1487
- [x] get_num_threads	1477
- [ ] real	1459

PR #144, opened 2021-09-27
Added cross batching rule

vfdev-5:
Description:
- Added cross batching rule
- Updated tests
- Fixed issue in test/common_utils.py with repeated number 3 in dims. Set it to 4 and 2 to avoid special value for cross op

Related to #101


Chillee:
uh.... does this not work if the batch size is 3?

vfdev-5:
> uh.... does this not work if the batch size is 3?

Seems like that :( 
If cross' `dim` is not specified then batch_size=3 will be found as the one to use.

Chillee:
Add another test (other than OpInfo) to cover the bs=3 case and we're good to go :)

Issue #112, opened 2021-09-08
Most used `torch.<x>` operations: 150-200


Chillee:
- [ ] quantize_per_tensor
- [ ] diag_embed
- [ ] lerp
- [ ] asin
- [ ] eig
- [ ] trunc
- [ ] diagonal
- [ ] cosh
- [ ] rfft
- [ ] cumprod
- [ ] addr
- [ ] roll
- [ ] narrow
- [ ] digamma
- [ ] square
- [ ] sinh
- [ ] logspace
- [ ] broadcast_tensors
- [ ] irfft
- [ ] frac
- [ ] hann_window
- [ ] solve
- [ ] logdet
- [ ] expm1
- [ ] cdist
- [ ] addmv
- [ ] randint_like
- [ ] tensordot
- [ ] ifft
- [ ] true_divide
- [ ] erfinv
- [ ] addcdiv
- [ ] addbmm
- [ ] renorm
- [ ] pinverse
- [ ] isclose
- [ ] erfc
- [ ] is_storage
- [ ] triangular_solve
- [ ] rot90
- [ ] logical_not
- [ ] geqrf
- [ ] slogdet
- [ ] lu
- [ ] hamming_window
- [ ] orgqr
- [ ] ormqr
- [ ] is_floating_point
- [ ] diagflat
- [ ] cholesky_solve

PR #145, opened 2021-09-27
Added roll batching rule

vfdev-5:
Description:
- Added roll batching rule
- Updated tests

Related to #112


functorch/csrc/BatchRulesViews.cpp
```diff
@@ -339,6 +350,7 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   VMAP_SUPPORT("squeeze", squeeze_batch_rule);
   VMAP_SUPPORT("squeeze.dim", squeeze_dim_batch_rule);
   VMAP_SUPPORT("_reshape_alias", _reshape_alias_batch_rule);
+  VMAP_SUPPORT("roll", roll_batch_rule);
```
Chillee, line 353: err, is this another one that can be solved by `REDUCTION_BOXED_ARGS`?

vfdev-5: Naively using `REDUCTION_BOXED_ARGS(roll, 2);` does not work and gives: `RuntimeError: shifts and dimensions must align. shifts: 1, dims:3`

Chillee: i think the issue comes when we're calling the equivalent of `torch.roll(x, shifts=1, dims=[])`.

The typical behavior for the boxed fallback is to expand `dims` to the entire array.

For example, `torch.sum(x: rank 3 tensor, dim=[])` => `torch.sum(x, dim=[0,1,2])`.

In this case, however, that behavior doesn't work.

So I guess we can't use `REDUCTION_BOXED_ARGS` :)

However, I think there's a case this PR isn't handling correctly either.

```
from functorch import vmap
import torch

def f(x):
    return torch.roll(x, 1)

val = torch.arange(3*4*5).view(3,4,5)
print(f(val[0]))
print(vmap(f)(val)[0])

>>>
tensor([[19,  0,  1,  2,  3],
        [ 4,  5,  6,  7,  8],
        [ 9, 10, 11, 12, 13],
        [14, 15, 16, 17, 18]])
tensor([[59,  0,  1,  2,  3],
        [ 4,  5,  6,  7,  8],
        [ 9, 10, 11, 12, 13],
        [14, 15, 16, 17, 18]])
```

vfdev-5: I see, thanks a lot for catching that ! I'll figure out how to cover that case (when dims is empty) and update the PR

Chillee (approved):

PR #147, opened 2021-09-27
Support buffers in compiled_module

bertmaher:


Chillee:
I think the main thing that's missing from this (and this probably requires some more infrastructural work...) is that currently `compiled_function` assumes you want to take the gradient of everything wrt its inputs. 

Unfortunately, we do not want to take the gradients wrt our buffers (most of the time!). For example, `batch_norm` stores an integer buffer that counts how many batches have gone through it.

test/test_pythonkey.py
```diff
@@ -351,6 +356,9 @@ def get_grads(args):
             for arg in args:
                 arg.requires_grad = True
 
+            # Recompile since we're changing which inputs requires grads.
+            compiled_f = compiled_function(f, lambda x,_: x, lambda x,_: x)
```
Chillee, line 360: The purpose of these tests is to make sure the function we're tracing isn't overspecializing on inputs.

Can you replace this with
```
            def create_new_arg(x):
                if x.requires_grad:
                    return x.detach().uniform_(0, 1).requires_grad_()
                else:
                    return x.detach().uniform_(0, 1)
            args = pytree.tree_map(create_new_arg, args)
```

I think that also fixes some of the previous failures :P 

Alternately, I can just merge this first and then change it.

Chillee (approved):
Can you add one more test tracing resnet18 or some other model to ensure the values/gradients are correct?

I'm imagining something like
```
model = models.densenet121()
inp = torch.randn(3,3,224,224)
model = compiled_module(model, lambda x,_: x, lambda x,_: x)
model(inp).sum().backward()
sm = 0
for p in model.parameters():
    sm += p.sum()
print(sm)
```

Chillee (approved):
Can you add one more test tracing resnet18 or some other model to ensure the values/gradients are correct?

I'm imagining something like
```
model = models.densenet121()
inp = torch.randn(3,3,224,224)
model = compiled_module(model, lambda x,_: x, lambda x,_: x)
model(inp).sum().backward()
sm = 0
for p in model.parameters():
    sm += p.sum()
print(sm)
```

bertmaher:
Cool, yes, I'll add a tracing test and switch to the non-recompiling version of the tests.  One minute!

PR #160, opened 2021-09-30
{log|softmax}_backward update as per the new signature

kshitij12345:
functorch was failing when trying to run on the latest PyTorch.

This [PR](https://github.com/pytorch/pytorch/pull/65242) changed the signature of `softmax` and `logsoftmax` backward functions and hence there was a failure while Registering.

I think this will pass once the nightly with changes from this PR is pulled by the CI.

### DIFF
--- a/functorch/csrc/OutOfPlacePlumbing.cpp
+++ b/functorch/csrc/OutOfPlacePlumbing.cpp
@@ -1643,11 +1643,11 @@ Tensor lowerToNextLayer<batch_rule_75_t,Tensor,const Tensor &, Dimname, c10::opt
   return makeBatched(std::get<0>(results), std::get<1>(results), cur_level);
 }
 
-typedef std::tuple<Tensor,c10::optional<int64_t>> (*batch_rule_76_t)(const Tensor &, c10::optional<int64_t>, const Tensor &, c10::optional<int64_t>, int64_t, const Tensor &, c10::optional<int64_t>);
+typedef std::tuple<Tensor,c10::optional<int64_t>> (*batch_rule_76_t)(const Tensor &, c10::optional<int64_t>, const Tensor &, c10::optional<int64_t>, int64_t, ScalarType);
 template <>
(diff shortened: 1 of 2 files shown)


functorch/csrc/OutOfPlacePlumbing.cpp
```diff
@@ -1643,11 +1643,11 @@ Tensor lowerToNextLayer<batch_rule_75_t,Tensor,const Tensor &, Dimname, c10::opt
   return makeBatched(std::get<0>(results), std::get<1>(results), cur_level);
 }
 
-typedef std::tuple<Tensor,c10::optional<int64_t>> (*batch_rule_76_t)(const Tensor &, c10::optional<int64_t>, const Tensor &, c10::optional<int64_t>, int64_t, const Tensor &, c10::optional<int64_t>);
+typedef std::tuple<Tensor,c10::optional<int64_t>> (*batch_rule_76_t)(const Tensor &, c10::optional<int64_t>, const Tensor &, c10::optional<int64_t>, int64_t, ScalarType);
```
kshitij12345, line 1646: Would be nice if these rules also mentioned the list/type of operators they correspond to 🤔

Chillee: These are codegen'ed, actually. Not sure if you modified it manually :P 

I do agree that it would be nice if they mentioned the operator lol.

kshitij12345: Ouch. Not sure why it didn't work for me. I cleaned the previous build and built it again but it didn't work (undefined symbol for this instantiation). Had to do it manually and it worked.

Chillee: Well, ideally, we'd integrate it into our build system... But that's kinda hard/annoying, and has some other ramifications too...
 
so.... we just haven't done it :P

Chillee:
LGTM!

soulitzer:
Had to revert this actually, so functorch might fail again, but will reland shortly.

PR #161, opened 2021-09-30
[port] `expand` to new api

kshitij12345:
Currently just making sure that the CI passes

TODO:
* [x] Add comments


kshitij12345:
cc: @zou3519 for review :)

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -342,6 +342,40 @@ std::tuple<Tensor,optional<int64_t>> slice_backward_batch_rule(
...
+              "expand: the number of sizes provided (", size.size(), ") ",
+              "must be greater or equal to the number of dimensions in the tensor (", static_cast<uint64_t>(self_dim - 1), ")");
+
+  auto self_ = moveBatchDimToFront(self, self_bdim);
+  auto self_sizes = self_.sizes();
+  auto batch_size = self_sizes[0];
+
+  c10::SmallBuffer<int64_t, 5> size_(size.size() + 1);
+  size_[0] = batch_size;
+  std::copy(size.cbegin(), size.cend(), size_.begin() + 1);
```
Chillee, line 359: I think you can just insert into the beginning of size here, right?

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -342,6 +342,40 @@ std::tuple<Tensor,optional<int64_t>> slice_backward_batch_rule(
...
+  // Here, we know we are expanding a (logical) tensor to a larger number
+  // of dimensions. We have to be careful because we can't call expand directly
+  // due to the presence of batch dimensions.
+  //
+  // As an example, let B0 be a batch dimension and consider expand(Tensor[B0, 3], [2, 3]).
+  // The result should be a tensor of size [B0, 2, 3].
+  // A physical view of size [B0, 3] can't directly be expanded to size [B0, 2, 3]
+  // so the strategy here is to view it first as a tensor of size [B0, 1, 3] and
+  // then expand.
+  auto extra_dims = size.size() - (self_dim - 1);
```
Chillee, line 370: I think this is just always guaranteed to be 1 now? Since we only have a single batch dimension on `self`, and since it's the only tensor it's guaranteed that we have a batch dimension.

Chillee: Similarly, I think some of the rest of the logic can be simplified - i.e. I think we can just insert again instead of copying.

Chillee (approved):
Mostly LGTM, I since there are some things we can clean up with the new API.

functorch/csrc/BatchRulesViews.cpp
zou3519: Inserting batch_size into the beginning of a sizes vector results in a shift of the data of rest of the sizes vector. It's a bit faster as it is currently written (we write the batch size first and then write the rest of the sizes)

functorch/csrc/BatchRulesViews.cpp
zou3519: One can do:
```
x = torch.randn(N)
vmap(lambda x: x.expand(2, 3, 4, 5))(x)
```
so `extra_dims` isn't always 1, right? In this case, it looks like 4

Chillee: ah right, misunderstood what the variables were.

PR #163, opened 2021-10-01
Python pointwise compiler implementation

bertmaher:
This is the python component of the op-authoring-with-NNC/pointwise compiler prototype.  I've done nothing here except move it from core to functorch, so likely none of this runs as-is.  It's here purely so someone can tidy it up and land it properly!


anijain2305:
Will work on it.

bertmaher:
This works now (modulo backwards, which I haven't added to the compilation cache just yet).

PR #164, opened 2021-10-01
[port] `view` to new api

kshitij12345:
port `view` to new api

TODO:
* [x] Add comment about porting this implementation.


vfdev-5:
@kshitij12345 I was working on porting it here: https://github.com/facebookresearch/functorch/pull/137 and it can be more trickier that it seems like.

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -342,6 +342,19 @@ std::tuple<Tensor,optional<int64_t>> slice_backward_batch_rule(
   return std::make_tuple(std::move(result), 0);
 }
 
+std::tuple<Tensor, optional<int64_t>> view_batching_rule(const Tensor& self, optional<int64_t> self_bdim, IntArrayRef size) {
+  auto self_ = moveBatchDimToFront(self, self_bdim);
```
vfdev-5, line 346: I think if you move to front it wont be a view

kshitij12345: @vfdev-5 Thanks for looking at it.
I thought the bdim are mostly hidden from users so it is ok to shuffle around (which should keep it as a view)

Since I was porting it, I thought the tests would cover the cases. How did you figure that #137 was wrong? #156 does not mention what was wrong about that implementation.

kshitij12345: I think it does return view

```python
import torch
import functorch

t = torch.randn(3, 9, 3)

def fn(x, sizes):
    return x.view(sizes)

def assert_is_view(t1, t2):
    assert t1.storage().data_ptr() == t2.storage().data_ptr()

assert_is_view(functorch.vmap(fn, in_dims=(0, None))(t, (27, 1)), t)
assert_is_view(functorch.vmap(fn, in_dims=(1, None))(t, (1, 1, 3, 3)), t)
assert_is_view(functorch.vmap(fn, in_dims=(2, None))(t, (1, 3, 9)), t)

```

vfdev-5: Well, in my PR I removed few tests that should be present: https://github.com/facebookresearch/functorch/pull/137/files#diff-ee63347c8ea255a43be0bd8e3c7b46c025877f06f3ab22538e409a5c4daac678L2436-L2437

Yes, seems like movedim (inside moveBatchDimToFront) returns a view but non-contiguous... I do not know if it is expected by user...

I think what we need to do is to insert bdim where it should be appropriate. For example, input tensor is [d1, d2, ... bdim,  dk, ... dN] and view shape is [D1, ... DN]. We have to insert bdim where it is appropriate without tweaking strides

kshitij12345: I am not sure if there is one correct output 🤔 

For Input Shape (9, B0, 3) -> View Shape (3, 1, 1, 3, 3), are we expecting (3, 1, 1, 3, B0, 3)? I think as long as we track the Batch dim correctly (it should be ok).

@zou3519 @Chillee what is your opinion on the same?

vfdev-5: I think we can figure out how to keep the stride for bdim instead of moving it somewhere

kshitij12345:
@vfdev-5 @Chillee @zou3519 
I think for the scope of this porting PR, the current implementation should be ok.

1. The current implementation also has same behaviour.
2. `reference_vmap` in `test_vmap` also moves the `bdim` to the front.
3. Implementation to preserve the position of `bdim` seems to be non-trivial. I tried to do it at #177

So for now I think it is ok to just port the implementation. Let me know if it sounds ok 
Thanks!

For 2.
```python
print(reference_vmap(torch.Tensor.view, (torch.randn(1, 2, 3, 4), (2, 4)), (2, None)).shape)
# prints torch.Size([3, 2, 4])
```

zou3519 (approved):
Port looks good. It's pretty difficult to get view's batching rule to error out correctly without moving the batch dim to the front (I don't know why this is); that's something we can punt to the future

PR #171, opened 2021-10-04
add cosine_similairity batching rule

kshitij12345:
Reference: https://github.com/facebookresearch/functorch/issues/102

### DIFF
--- a/functorch/csrc/BatchRulesBinaryOps.cpp
+++ b/functorch/csrc/BatchRulesBinaryOps.cpp
@@ -178,8 +178,10 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   POINTWISE_BOXED(clamp.Tensor);
   BINARY_POINTWISE2(clamp_min, Tensor);
   UNARY_POINTWISE(clamp_min);
+  POINTWISE_BOXED(clamp_min_);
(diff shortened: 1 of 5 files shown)


Chillee:
So, `cosine_similarity` is actually a composite op (you can tell since t's marked as Composite here: https://docs.google.com/spreadsheets/d/1Sp4HUjxwMifS5oDQg0yvjqk7hKOpCfKO4jWH4MTGP-k/edit#gid=0) and also, it's in `STOP_DECOMPOSITION` :P)

So, in this case, we don't actually need to write a batching rule for `aten::cosine_similarity`, we can just write one for the composites. You can tell what the composites are by either running it individually with vmap (and viewing the warnings about falling back), or by running `pytest -k cosine_similairty`. In this case, I think we can implement the batching rule for `cosine_similarity` by implementing the batching rule for `aten::clamp_min_` (which you can do with `POINTWISE_BOXED`), and then removing the appropriate `STOP_DECOMPOSITION` line.

PS: Sometimes, even though an op is composite, it may have code inside of it (such as certain kinds of in-place ops or control flow) that causes vmap to break. Those cases are a bit trickier, and you can tell you're in one of these cases if commenting out `STOP_DECOMPOSITION` makes vmapping that op break.

Chillee (approved):

Chillee:
LGTM! Thanks.

PR #173, opened 2021-10-04
Added tests for op aliases

vfdev-5:


functorch/csrc/BatchRulesStopDecomposition.cpp
```diff
@@ -438,12 +425,11 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   STOP_DECOMPOSE(mode.dimname);
   STOP_DECOMPOSE(mode.dimname_out);
   STOP_DECOMPOSE(moveaxis.int);
```
zou3519, line 427: This is also an alias, right?

functorch/csrc/BatchRulesStopDecomposition.cpp
```diff
@@ -438,12 +425,11 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
...
-  STOP_DECOMPOSE(moveaxis.intlist);
+  // STOP_DECOMPOSE(moveaxis.intlist);
   STOP_DECOMPOSE(msort.out);
   STOP_DECOMPOSE(multilabel_margin_loss);
   STOP_DECOMPOSE(multilabel_margin_loss.out);
   STOP_DECOMPOSE(multiply.Scalar);
```
zou3519, line 432: This is also an alias

functorch/csrc/BatchRulesStopDecomposition.cpp
```diff
@@ -474,7 +460,6 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   STOP_DECOMPOSE(normal.float_float_out);
   STOP_DECOMPOSE(not_equal.Scalar);
```
zou3519, line 461: Also alias

functorch/csrc/BatchRulesStopDecomposition.cpp
```diff
@@ -615,17 +596,14 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   STOP_DECOMPOSE(stride.Dimname);
   STOP_DECOMPOSE(stride.int);
   STOP_DECOMPOSE(subtract.Scalar);
```
zou3519, line 598: Also alias

functorch/csrc/BatchRulesStopDecomposition.cpp
vfdev-5: Yes, they are. I just removed those that are tested. Others does not fail and thus are not tested.

zou3519 (commented):
Approach is reasonable. We only really need tests for `vmap` for these transforms; vjp always decomposes the operation because of how autograd works and because of how aliases work as long as all of our transforms successfully decompose the alias then everything else should work out.

I noticed that in this PR there are tests for vmapvjp, vjpvmap, and vmap, so we should have enough test coverage for these.

I added some suggestions for other aliases we can remove from the stop_decompose, but other than that this looks pretty good.

functorch/csrc/BatchRulesStopDecomposition.cpp
zou3519: Interesting. So a lot of the `foo.Scalar` overloads cannot be called directly from Python, but can be hit from C++ (yes, it's weird). We should still remove the stop_decompose because we're pretty sure that they work.

functorch/csrc/BatchRulesStopDecomposition.cpp
```diff
@@ -207,9 +200,7 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   STOP_DECOMPOSE(diagflat);
   STOP_DECOMPOSE(diagonal.Dimname);
   STOP_DECOMPOSE(diff.out);
-  STOP_DECOMPOSE(divide.Scalar);
   STOP_DECOMPOSE(divide.Scalar_mode);
```
vfdev-5, line 203: divide.Scalar_mode and divide.Tensor_mode are also aliases, should we remove them as well ?

zou3519: Yup, go for it

zou3519 (approved):

PR #180, opened 2021-10-05
add aminmax batching rule

kshitij12345:
`aminmax` has a different behavior compared to other reduction ops for scalar tensor.

```python
t = torch.randn(())

result = torch.aminmax(t, dim=0, keepdim=True)
print(result[0].shape, result[1].shape)

result = torch.sum(t, dim=0, keepdim=True)
print(result.shape)
```

Output
```
torch.Size([1]) torch.Size([1])
torch.Size([])
```

Due to this we can't use `REDUCTION_BOXED_ARGS` as is (due to the following lines).
https://github.com/facebookresearch/functorch/blob/17f407c9faea3ec5f953b9afb17d7c8c7e6f34fb/functorch/csrc/BatchRulesReduceOps.cpp#L162-L164

For this reason, have added a batching rule. Other option would be to update `REDUCTION_BOXED_ARGS` to accommodate this case.

### DIFF
--- a/functorch/csrc/BatchRulesReduceOps.cpp
+++ b/functorch/csrc/BatchRulesReduceOps.cpp
@@ -281,6 +281,13 @@ std::tuple<Tensor, optional<int64_t>, Tensor, optional<int64_t>> aminmax_batchin
   }
 
   std::tie(min, max) = at::aminmax(self_, dim, keep_dim);
+
+  if (logical_rank == 0 && self_.device().is_cuda()) {
+    // behaviour diverges between cpu and cuda
+    // reference: https://github.com/pytorch/pytorch/issues/64008
+    min = min.squeeze(-1);
+    max = max.squeeze(-1);
+  }
   return std::make_tuple(min, 0, max, 0);
 }
 
--- a/functorch/csrc/BatchRulesReduceOps.cpp
+++ b/functorch/csrc/BatchRulesReduceOps.cpp
@@ -255,16 +255,12 @@ std::tuple<Tensor,optional<int64_t>> _log_softmax_backward_batch_rule(
   return std::make_tuple(at::_log_softmax_backward_data(grad_output_, output_, dim, input_dtype), 0);
 }
 
+// aminmax has divergent behavior for 0-d tenosrs.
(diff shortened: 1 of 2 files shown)


Chillee:
I think this is a bug, btw (on the `aminmax` side): https://github.com/pytorch/pytorch/issues/64008

I think it's good to land this as is, but we might want to change it eventually.

kshitij12345:
> I think this is a bug, btw (on the aminmax side): pytorch/pytorch#64008

Ah. Right. Seemed odd. Thanks for confirming!

functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -255,6 +255,42 @@ std::tuple<Tensor,optional<int64_t>> _log_softmax_backward_batch_rule(
   return std::make_tuple(at::_log_softmax_backward_data(grad_output_, output_, dim, input_dtype), 0);
 }
 
+std::tuple<Tensor, optional<int64_t>, Tensor, optional<int64_t>> aminmax_batching_rule(
+    const Tensor &self, optional<int64_t> self_bdim, optional<int64_t> dim, bool keep_dim)
+{
+  Tensor min, max;
+  if (!self_bdim.has_value())
```
Chillee, line 262: I think this case never happens now (if no input tensors have batch dims we don't dispatch to the batch rule)

functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -255,6 +255,42 @@ std::tuple<Tensor,optional<int64_t>> _log_softmax_backward_batch_rule(
...
+    // flatten the input except for batch-dim
+    auto bsize = self.size(0);
+    self_ = self.view({bsize, -1});
+    dim = 1;
+  }
+
+  std::tie(min, max) = at::aminmax(self_, dim, keep_dim);
+
+  if (logical_rank == 0 && self_.device().is_cuda()) {
+    // behaviour diverges between cpu and cuda
```
Chillee, line 286: :O

I didn't realize behavior diverged lol.

PR #187, opened 2021-10-06
Fix CI by unpinning the binary version

zou3519:


zou3519:
Well, at least the build succeeds

PR #211, opened 2021-10-17
convert movedim to new api

Padarn:
PR converts `movedim` to the new batching rule API.

Very new to the project so please don't hesitate to correct anything where I've broken style/naming conventions etc.

Thanks!


Padarn:
The test case failures in circleci look unrelated to my change, but I'm not confident. Are these expected currently?

Padarn:
@Chillee are the builds that were failing on main something I can help resolve? or the problem is known?

Issue #224, opened 2021-10-25
functorch preview branch checklist


zou3519:
Add and audit docstrings for APIs
- [x] vmap @zou3519 
- [x] grad (#231), grad_and_value @samdow 
- [x] vjp @samdow 
- [x] jacrev @samdow 
- [x] make_fx @zou3519 (resolution: I don't think I'm going to document this)
- [x] vjpfull (Confirm if this should be exposed. I don't think it should be) @zou3519 
- [x] make_functional, make_functional_with_buffers, FunctionalModule @zou3519 
- [x] combine_state_for_ensemble @zou3519 
- [x] Figure out if {make_functional, make_functional_with_buffers, FunctionalModule, combine_state_for_ensemble} need their own namespace. @zou3519 
- [x] Should we expose all of {wrap_key, PythonTensor, pythonkey_trace, make_fx}? Hypothesis: we only need to expose make_fx  @zou3519 
- [x] all compile things go into functorch.compile @zou3519 
- [x] Excise functional_init* @zou3519 

Do we need a docs website?
- [x] Answer the question
- [x] Make the site @zou3519

Installation instructions
- [x] Add installation instructions for PyTorch 1.10 to README / website. Set up colab notebook with PyTorch 1.10 installation. @zou3519

Misc
- [x] delete the deepcopy in make_functional @zou3519 
- [x] delete make_functional_deprecated_v1, make_functional_with_buffers_deprecated_v1 (it's been long enough)
- [x] Perform cherry-picks of all of the above into the 1.10 branch @zou3519 
- [x] Push docs website (functorch.github.io) and link from readme @zou3519 
- [ ] Marketing (tweets) @Chillee

zou3519:
One item left!

zou3519:
We did the preview branch

PR #231, opened 2021-10-28
Added docstring for `grad`

vfdev-5:
Description:
- Addresses #224
- Added docs for grad. 
- Cleaned unused imports and other nits

### DIFF
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -4,20 +4,15 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from typing import Callable, Union, Tuple
 import torch
 from functools import partial, wraps
 import contextlib
-import collections
-import torch.nn as nn
-import torch.nn.functional as F
 from torch.utils._pytree import tree_flatten, tree_unflatten, tree_map
 from .pytree_hacks import tree_map_, treespec_pprint
-from collections import namedtuple
 import torch.autograd.forward_ad as fwAD
-import gc
 
 from .vmap import vmap
-from .make_functional import make_functional_deprecated_v1, make_functional_with_buffers_deprecated_v1
 
 from functorch._C import (
     _wrap_for_grad,
@@ -25,6 +20,9 @@ from functorch._C import (
     _grad_increment_nesting,
     _grad_decrement_nesting,
 )
+
+argnums_t = Union[int, Tuple[int, ...]]
+
 
 def _create_differentiable(inps, level=None):
     def create_differentiable(x):
@@ -278,8 +276,8 @@ def jacfwd(f):
         return result
     return wrapper_fn
 
-def grad_and_value(f, argnums=0, has_aux=False):
-    @wraps(f)
+def grad_and_value(func: Callable, argnums: argnums_t = 0, has_aux: bool = False) -> Callable:
+    @wraps(func)
     def wrapper(*args, **kwargs):
         level = _grad_increment_nesting()
         output, aux, grad_input = None, None, None
@@ -291,7 +289,7 @@ def grad_and_value(f, argnums=0, has_aux=False):
                 diff_args = _slice_argnums(args, argnums)
                 tree_map_(partial(_create_differentiable, level=level), diff_args)
 
-                output = f(*args, **kwargs)
+                output = func(*args, **kwargs)
                 if has_aux:
                     output, aux = output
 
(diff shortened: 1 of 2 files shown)


vfdev-5:
@samdow sounds good !

> >  By the way we can omit typehints in docstring and use a sphinx plugin which would get the typehint from the code

> Noticed that you did this for grad and this is great!

Actually, I did put typehints to the docstring. For example

```python
func (Callable): A Python function that takes one or more arguments.
```

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        In this case, ``grad(f)(x)`` will respect the inner ``torch.no_grad``.
+
+        Case 2: Using ``grad`` inside ``torch.no_grad`` context manager:
+
+            >>> with torch.no_grad():
+            >>>     grad(f)(x)
+
+        In this case, ``grad`` will respect the inner ``torch.no_grad``, but not the
+        outer one. This is because ``grad`` is a "function transform": its result
+        should not depend on the result of a context manager outside of ``f``.
```
samdow, line 398: Thanks for copying this note up!

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+    with respect to the input. This operator can be used multiple times to compute
+    higher-order gradients.
+
+    Args:
+        func (Callable): A Python function that takes one or more arguments.
+            Must return a single-element Tensor. If specified ``has_aux=True``, function can
+            return a tuple of single-element Tensor and other auxiliairy objects:
+            ``(output, aux)``.
+        argnums (int or Tuple[int]): Specifies arguments to compute gradients with respect to.
+            ``argnums`` can be single integer or tuple of integers. Default: 0.
```
samdow, line 337: nit:
```suggestion
            :attr:`argnums` can be single integer or tuple of integers. Default: 0.
```

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        has_aux (bool): Flag indicating that ``func`` returns a tensor and other
```
samdow, line 338: nit: 
```suggestion
        has_aux (bool): Flag indicating that :attr:`func` returns a tensor and other
```

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+            auxiliairy objects: ``(output, aux)``. Default: False.
+
+    Returns:
+        Function to compute gradients on its inputs. By default, the output of
+        the function is gradients tensor. If specified ``has_aux=True``, tuple of gradients
+        and output auxiliairy objects is returned.
+        If ``argnums`` is a tuple of integers, a tuple of output gradients with respect to
```
samdow, line 345: nit: 
```suggestion
        If :attr:`argnums` is a tuple of integers, a tuple of output gradients with respect to
```

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        each ``argnums`` value is returned
```
samdow, line 346: nit:
```suggestion
        each :attr:`argnums` value is returned
```

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+    Example of using ``grad``:
+
+        >>> from functorch import grad
+        >>> x = torch.randn([])
+        >>> cos_x = grad(lambda x: torch.sin(x))(x)
+        >>> assert torch.allclose(cos_x, x.cos())
+        >>>
+        >>> # Second-order gradients
+        >>> neg_sin_x = grad(grad(lambda x: torch.sin(x)))(x)
+        >>> assert torch.allclose(neg_sin_x, -x.sin())
```
samdow, line 357: Thanks for including the second order gradient example!

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        >>> def compute_loss(weights, example, target):
+        >>>     y = model(weights, example)
+        >>>     return ((y - target) ** 2).mean()  # MSELoss
+        >>>
+        >>> weights = torch.randn(feature_size, requires_grad=True)
+        >>> examples = torch.randn(batch_size, feature_size)
+        >>> targets = torch.randn(batch_size)
+        >>> inputs = (weights,examples, targets)
+        >>> grad_weight_per_example = vmap(grad(compute_loss), in_dims=(None, 0, 0))(*inputs)
+
```
samdow, line 378: Could we add an example demonstrating the `has_aux` flag? Like `in_dims`, the example can be more contrived than the other examples if need be just so a new user can see using that flag

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        >>>     return feature_vec.dot(weights).relu()
+        >>>
+        >>> def compute_loss(weights, example, target):
+        >>>     y = model(weights, example)
+        >>>     return ((y - target) ** 2).mean()  # MSELoss
+        >>>
+        >>> weights = torch.randn(feature_size, requires_grad=True)
+        >>> examples = torch.randn(batch_size, feature_size)
+        >>> targets = torch.randn(batch_size)
+        >>> inputs = (weights,examples, targets)
```
samdow, line 376: nit: missing space after weight
```suggestion
        >>> inputs = (weights, examples, targets)
```

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        >>> # Second-order gradients
+        >>> neg_sin_x = grad(grad(lambda x: torch.sin(x)))(x)
+        >>> assert torch.allclose(neg_sin_x, -x.sin())
+
+    When composed with ``vmap``, ``grad`` can be used to compute per-sample-gradients:
+
+        >>> from functorch import vmap
+        >>> batch_size, feature_size = 3, 5
+        >>>
+        >>> def model(weights,feature_vec):
```
samdow, line 364: nit: space in between parameters
```suggestion
        >>> def model(weights, feature_vec):
```

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        >>> from functorch import grad
+        >>> x = torch.randn([])
+        >>> cos_x = grad(lambda x: torch.sin(x))(x)
+        >>> assert torch.allclose(cos_x, x.cos())
+        >>>
+        >>> # Second-order gradients
+        >>> neg_sin_x = grad(grad(lambda x: torch.sin(x)))(x)
+        >>> assert torch.allclose(neg_sin_x, -x.sin())
+
+    When composed with ``vmap``, ``grad`` can be used to compute per-sample-gradients:
```
samdow, line 359: This is a really great example--it feels like something users can immediately pull down and start modifying and building off of.

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        >>>
+        >>> def compute_loss(weights, example, target):
+        >>>     y = model(weights, example)
+        >>>     return ((y - target) ** 2).mean()  # MSELoss
+        >>>
+        >>> weights = torch.randn(feature_size, requires_grad=True)
+        >>> examples = torch.randn(batch_size, feature_size)
+        >>> targets = torch.randn(batch_size)
+        >>> inputs = (weights,examples, targets)
+        >>> grad_weight_per_example = vmap(grad(compute_loss), in_dims=(None, 0, 0))(*inputs)
```
samdow, line 377: Could we either add another example or adjust this one to demonstrate using `in_dims`? Like `has_aux`, the example can be more contrived than the other examples if there's no good one just so a new user can see how it's used

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+            Must return a single-element Tensor. If specified ``has_aux=True``, function can
+            return a tuple of single-element Tensor and other auxiliairy objects:
+            ``(output, aux)``.
+        argnums (int or Tuple[int]): Specifies arguments to compute gradients with respect to.
+            ``argnums`` can be single integer or tuple of integers. Default: 0.
+        has_aux (bool): Flag indicating that ``func`` returns a tensor and other
+            auxiliairy objects: ``(output, aux)``. Default: False.
+
+    Returns:
+        Function to compute gradients on its inputs. By default, the output of
```
samdow, line 342: nit:
```suggestion
        Function to compute gradients with respect to its inputs. By default, the output of
```

Also, although we've mentioned it before, it would be good to note that the returned function takes the same inputs as the original in case a user misses it once

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
...
+        the function is gradients tensor. If specified ``has_aux=True``, tuple of gradients
```
samdow, line 343: Maybe
```suggestion
        the function is the gradient tensor with respect to the first argument If specified ``has_aux=True``, tuple of gradients
```

samdow (approved):
This looks really great. I had some wording suggestions (glad to discuss if you disagree with any) + nits (it seems from the PyTorch docs that `` :attr:`<attr name>` `` is preferred over ` ``<attr name>`` ` ). I also requested a couple more examples, but if you don't want to add those, we can add them in a separate PR

vfdev-5:
Thanks for the review @samdow , I agree with your suggestions! Examples I added here are adaptations from readme and comments in the code. I'll add few others as suggested.

functorch/_src/eager_transforms.py
```diff
@@ -324,14 +322,89 @@ def wrapper(*args, **kwargs):
         return grad_input, output
     return wrapper
 
-def grad(f, argnums=0, has_aux=False):
-    @wraps(f)
+def grad(func: Callable, argnums: argnums_t = 0, has_aux: bool = False) -> Callable:
+    """``grad`` operator helps computing gradients of the output of a function.
```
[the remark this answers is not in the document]
vfdev-5: I agree, maybe we can merge this and below sentences in a single better one.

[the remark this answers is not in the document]
samdow: Agreed! Let me know if you want help wordsmithing it

functorch/_src/eager_transforms.py
vfdev-5: Added an example of using `has_aux` and `argnums` together. @samdow what do you think?

samdow: Looks awesome! Thanks for adding it!

samdow (approved):
Thanks for all the updates! Looks great

vfdev-5:
@samdow can we merge this PR or do you want me to add something else here ?

samdow:
Sorry I didn't realize that I had to merge it. Thanks again for adding this!

vfdev-5:
I do not have (yet) write rights for this repository :) that's why I asked you.

PR #232, opened 2021-10-29
Add per-op tests for jvp and vmap + jvp

soulitzer:
This PR depends on some fixes in core to land, so we expect CI to fail here for now

This PR:
- Updates `jvp` to support multiple inputs and outputs
- Fixes batched tensor storage offset to be the same as the wrapped tensor
- Adds per-op test for `jvp` by comparing with just normal forward AD in core
- Adds per-op test for `vmap(jvp` by comparing the looped version with batched version

cc @albanD @zou3519 @Chillee

### DIFF
--- a/functorch/csrc/BatchedTensorImpl.cpp
+++ b/functorch/csrc/BatchedTensorImpl.cpp
@@ -40,6 +40,7 @@ BatchedTensorImpl::BatchedTensorImpl(Tensor value, int64_t bdim, int64_t level)
     sizes_and_strides_.size_at_unchecked(dim) = value_sizes.at(actual_dim);
     sizes_and_strides_.stride_at_unchecked(dim) = value_strides.at(actual_dim);
   }
+  storage_offset_= value_.storage_offset();
   refresh_numel();
   refresh_contiguous();
 }
@@ -71,6 +72,7 @@ void BatchedTensorImpl::refreshSizesAndStrides() {
     sizes_and_strides_.size_at_unchecked(dim) = value_sizes.at(actual_dim);
     sizes_and_strides_.stride_at_unchecked(dim) = value_strides.at(actual_dim);
   }
+  storage_offset_= value_.storage_offset();
   refresh_numel();
   refresh_contiguous();
 }
--- a/functorch/csrc/BatchedTensorImpl.cpp
+++ b/functorch/csrc/BatchedTensorImpl.cpp
@@ -59,10 +59,10 @@ BatchedTensorImpl::BatchedTensorImpl(DispatchKeySet key_set, Tensor value, int64
   set_storage_access_should_throw();
   set_has_contiguity_policy(HasContiguityPolicy::CustomBehavior);
   checkInvariants();
-  refreshSizesAndStrides();
+  refreshTensorMetadata();
 }
 
-void BatchedTensorImpl::refreshSizesAndStrides() {
+void BatchedTensorImpl::refreshTensorMetadata() {
   const auto public_dims = value_.dim() - 1;
   const auto value_sizes = value_.sizes();
   const auto value_strides = value_.strides();
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -36,7 +36,9 @@ def loop(op, in_dims, out_dim, batch_size, *batched_args, **kwarg_values):
     return loop_out
 
 
-def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3, first_dim_only=False):
+def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3, bdims=(0, -1)):
+    assert bdims == (0,) or bdims == (0, -1)
+
     def add_batch_dim(arg, bdim, batch_size=3):
         assert bdim == 0 or bdim == -1
         if isinstance(arg, torch.Tensor):
@@ -50,7 +52,6 @@ def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3, first_
             assert False
         else:
             return (arg, None)
-    bdims = [0] if first_dim_only else [0, -1]
     for bdim in bdims:
         batch_choices = []
         def add_batch_choices(a):
@@ -73,10 +74,10 @@ def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3, first_
             yield pytree.tree_unflatten(batched_args, arg_spec), pytree.tree_unflatten(in_dims, arg_spec), kwarg_values
 
 
-def get_fallback_and_vmap_exhaustive(op, arg_values, kwarg_values, compute_loop_out=True, first_dim_only=False):
+def get_fallback_and_vmap_exhaustive(op, arg_values, kwarg_values, compute_loop_out=True, bdims=(0, -1)):
     out_dim = 0
     batch_size = 4
-    generator = get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size, first_dim_only=first_dim_only)
+    generator = get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size, bdims=bdims)
     for batched_args, in_dims, kwarg_values in generator:
         if compute_loop_out:
             loop_out = loop(op, in_dims, out_dim, batch_size, *batched_args, **kwarg_values)
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -26,6 +26,7 @@ def loop(op, in_dims, out_dim, batch_size, *batched_args, **kwarg_values):
         new_args = [a.select(in_dim, idx) if in_dim is not None else a for a, in_dim in zip(flat_args, flat_dims)]
         out = op(*pytree.tree_unflatten(new_args, args_spec), **kwarg_values)
         outs.append(out)
+
     loop_out = []
     if isinstance(outs[0], torch.Tensor):
         loop_out = torch.stack(outs)
@@ -35,7 +36,7 @@ def loop(op, in_dims, out_dim, batch_size, *batched_args, **kwarg_values):
     return loop_out
 
 
-def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3):
+def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3, first_dim_only=False):
     def add_batch_dim(arg, bdim, batch_size=3):
         assert bdim == 0 or bdim == -1
         if isinstance(arg, torch.Tensor):
@@ -49,8 +50,8 @@ def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3):
             assert False
         else:
             return (arg, None)
-
-    for bdim in [0, -1]:
+    bdims = [0] if first_dim_only else [0, -1]
+    for bdim in bdims:
         batch_choices = []
         def add_batch_choices(a):
             if isinstance(a, torch.Tensor):
@@ -72,10 +73,10 @@ def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3):
             yield pytree.tree_unflatten(batched_args, arg_spec), pytree.tree_unflatten(in_dims, arg_spec), kwarg_values
 
 
-def get_fallback_and_vmap_exhaustive(op, arg_values, kwarg_values, compute_loop_out=True):
+def get_fallback_and_vmap_exhaustive(op, arg_values, kwarg_values, compute_loop_out=True, first_dim_only=False):
     out_dim = 0
     batch_size = 4
-    generator = get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size)
+    generator = get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size, first_dim_only=first_dim_only)
     for batched_args, in_dims, kwarg_values in generator:
         if compute_loop_out:
             loop_out = loop(op, in_dims, out_dim, batch_size, *batched_args, **kwarg_values)
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -61,12 +61,12 @@ def _autograd_grad(outputs, inputs, grad_outputs=None, retain_graph=False, creat
     return tree_unflatten(result, inputs_spec)
 
 
-def diff_arg(arg, is_jvp=False):
+def diff_arg(arg, requires_grad=True):
     def is_differentiable_arg(arg):
-        if is_jvp:
-            return arg.is_floating_point() or arg.is_complex()
-        else:
+        if requires_grad:
             return arg.requires_grad
+        else:
+            return arg.is_floating_point() or arg.is_complex()
     if is_iterable_of_tensors(arg):
         if all([is_differentiable_arg(a) for a in arg]):
             return True
@@ -80,9 +80,9 @@ def diff_arg(arg, is_jvp=False):
 # - f' takes only positional arguments
 # - All arguments to f' are floating-point Tensors
 # - All outputs of f' are floating-point Tensors
-def normalize_op_input_output2(f, args, kwargs, output_process_fn_grad=None, is_jvp=False):
+def normalize_op_input_output2(f, args, kwargs, output_process_fn_grad=None, requires_grad=True):
     flat_args, args_spec = tree_flatten(args)
-    diff_argnums = tuple(i for i, arg in enumerate(flat_args) if diff_arg(arg, is_jvp=is_jvp))
+    diff_argnums = tuple(i for i, arg in enumerate(flat_args) if diff_arg(arg, requires_grad=requires_grad))
 
     assert len(diff_argnums) > 0
     primals = tuple(flat_args[i] for i in diff_argnums)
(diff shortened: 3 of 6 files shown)


zou3519:
Tests are failing, is this expected?

soulitzer:
Yup, the changes in core should address these failures

functorch/_src/eager_transforms.py
```diff
@@ -249,13 +249,10 @@ def jvp(f, primals, tangents):
...
         ctx = fwAD.dual_level if JVP_NESTING == 1 else noop
         with ctx():
-            # TODO: extend this to any number of primals
-            assert len(primals) == 1 and len(tangents) == 1
             duals = tuple(fwAD.make_dual(p, t) for p, t in zip(primals, tangents))
             result_duals = f(*duals)
             result_duals, _ = tree_flatten(result_duals)
-            assert len(result_duals) == 1
-            primals_out, tangents_out = fwAD.unpack_dual(result_duals[0])
+            primals_out, tangents_out = zip(*(fwAD.unpack_dual(d) for d in result_duals))
```
zou3519, line 255: You should pull main and use the changes from main; I did the exact same thing here and added an error check

test/test_ops.py
```diff
@@ -192,6 +228,48 @@ def wrapped_fn(*args, **kwargs):
...
+        xfail('minimum'),
+        xfail('min', 'binary'),
+        xfail('maximum'),
+        xfail('max', 'binary'),
+
+        # RuntimeError: Cannot access data pointer of Tensor that doesn't have storage
+        xfail('linalg.inv'),
+        xfail('linalg.matrix_power'),
+        xfail('linalg.cholesky'),
+        xfail('tensor_split'),
```
zou3519, line 247: These are Composite Compliance problems -- these operations are composite and they are not written in the correct way. I'll open an issue on PyTorch about them soon but they probably won't pass anytime soon

test/test_ops.py
```diff
@@ -192,6 +228,48 @@ def wrapped_fn(*args, **kwargs):
 
             self.assertEqual(result, expected)
 
+    @ops(functorch_lagging_op_db + additional_op_db, allowed_dtypes=(torch.float,))
+    @skipOps('TestOperators', 'test_jvp', set({
+        # Apprently these support forward AD, but we get "Trying to use forward AD..."
+        xfail('var_mean'),
+        xfail('std_mean'),
```
zou3519, line 235: Do they? I don't see "result: " entries in derivatives.yaml

https://github.com/pytorch/pytorch/blob/3e3bf40b0a5acc68db40d0c2570207f2f35ae557/tools/autograd/derivatives.yaml#L1524-L1526

test/test_ops.py
```diff
@@ -192,6 +228,48 @@ def wrapped_fn(*args, **kwargs):
...
+
+        # RuntimeError: expected scalar type double but found float
+        xfail('minimum'),
+        xfail('min', 'binary'),
+        xfail('maximum'),
+        xfail('max', 'binary'),
```
zou3519, line 241: I'm getting `NotImplementedError: Cannot access storage of TensorWrapper` on my branch after patching in your stack from https://github.com/pytorch/pytorch/pull/67041 and this PR, am I doing something wrong?

zou3519: Nevermind, the problem is I am building in debug mode and there are those autograd asserts...

test/test_ops.py
```diff
@@ -192,6 +228,48 @@ def wrapped_fn(*args, **kwargs):
 
             self.assertEqual(result, expected)
 
+    @ops(functorch_lagging_op_db + additional_op_db, allowed_dtypes=(torch.float,))
+    @skipOps('TestOperators', 'test_jvp', set({
+        # Apprently these support forward AD, but we get "Trying to use forward AD..."
+        xfail('var_mean'),
+        xfail('std_mean'),
+
+        # RuntimeError: expected scalar type double but found float
```
zou3519, line 237: I have some vague idea of what is wrong here: https://github.com/pytorch/functorch/issues/274

zou3519: Edit: nevermind

test/test_ops.py
```diff
@@ -192,6 +228,48 @@ def wrapped_fn(*args, **kwargs):
...
+        xfail('minimum'),
+        xfail('min', 'binary'),
+        xfail('maximum'),
```
zou3519, line 240: @soulitzer I dug into this a little. "expected scalar type double but found float" is a bug with PyTorch forward-mode AD, not functorch. We should leave a comment here to that effect (or fix the forward formula for these)
```
import torch
import torch.autograd.forward_ad as fwAD

x = torch.randn(3)
y = torch.randn(3)
tx = torch.randn(3)
ty= torch.randn(3)

with fwAD.dual_level():
  x_dual = fwAD.make_dual(x, tx)
  y_dual = fwAD.make_dual(y, ty)
  torch.maximum(x_dual, y_dual)
```

test/test_ops.py
zou3519:
```
import torch
import torch.autograd.forward_ad as fwAD

x = torch.randn(3)
y = torch.randn(3)

with fwAD.dual_level():
  x_dual = fwAD.make_dual(x, tx)
  torch.std_mean(x_dual)
```
gives me:
```
NotImplementedError: Trying to use forward AD with std_mean that does not support it.
```

test/test_ops.py
```diff
@@ -192,6 +228,48 @@ def wrapped_fn(*args, **kwargs):
...
+
+        # RuntimeError: Cannot access data pointer of Tensor that doesn't have storage
+        xfail('linalg.inv'),
+        xfail('linalg.matrix_power'),
+        xfail('linalg.cholesky'),
+        xfail('tensor_split'),
+
+        # NotImplementedError: Trying to use forward AD with aten::_new_zeros_with_same_meta
+        # that does not support it. (this is weird)
+        xfail('linalg.eigh'),
```
zou3519, line 251: Seems to pass locally for me now

test/test_ops.py
```diff
@@ -388,6 +466,101 @@ def test_vmapvjp(self, device, dtype, op):
...
+        xfail('nn.functional.dropout'),  # randomness
+        xfail('double'),
+        xfail('linalg.householder_product'),
+
+        # Batching rule not implemented for aten::_local_scalar_dense
+        xfail('linalg.eigh'),
+        xfail('linalg.tensorinv'),
+        xfail('linalg.inv'),
+        xfail('linalg.matrix_power'),
+        xfail('linalg.cholesky'),
```
zou3519, line 488: We should note that jvp and vmap composition in pytorch core also has this problem

test/test_ops.py
```diff
@@ -388,6 +466,101 @@ def test_vmapvjp(self, device, dtype, op):
...
+
+        # Batching rule not implemented for aten::logical_{and,or}.out
+        xfail('cumprod'),
+        xfail('__rpow__'),
+        xfail('clamp', 'scalar'),
+        xfail('logit'),
+        xfail('fmin'),
+        xfail('fmax'),
+        xfail('float_power'),
+        xfail('pow'),
```
zou3519, line 498: To summarize, I think the things that need looking into from the functorch side are the following categories:
```
# Random failures: Copied from above
...

# Expected the output of forward differentiable view operations
# to have the tangent have the same layout as primal
...

# RuntimeError: Cannot access data pointer of Tensor that doesn't have storage
...
```

test/test_ops.py
```diff
@@ -388,6 +466,101 @@ def test_vmapvjp(self, device, dtype, op):
...
+        xfail('linalg.householder_product'),
+
+        # Batching rule not implemented for aten::_local_scalar_dense
+        xfail('linalg.eigh'),
+        xfail('linalg.tensorinv'),
+        xfail('linalg.inv'),
+        xfail('linalg.matrix_power'),
+        xfail('linalg.cholesky'),
+
+        # Batching rule not implemented for aten::logical_{and,or}.out
```
zou3519, line 490: Are these problems in PyTorch core? (I would expect them to be)

test/test_ops.py
```diff
@@ -388,6 +466,101 @@ def test_vmapvjp(self, device, dtype, op):
...
+        xfail('index_put'),
+        xfail('index_fill'),
+        xfail('masked_fill'),
+        xfail('masked_scatter'),
+
+        # Expected the output of forward differentiable view operations
+        # to have the tangent have the same layout as primal
+        xfail('linalg.multi_dot'),
+        xfail('einsum'),
+        xfail('reshape'),  # this is odd
```
zou3519, line 518: I can't repro these failures -- are they CUDA-only? (i have a cpu-only build)

soulitzer: This is on CPU, you aren't able to replicate it because I tested this on top of the stack in core! See https://github.com/pytorch/pytorch/pull/66293

zou3519: Ah I see this now. I was not checking out your stack in core correct, thanks for the correction!

test/test_ops.py
soulitzer: lmao the OpInfo has `supports_forward_ad=True`, but grad, gradgrad checks are all disabled

test/test_ops.py
```diff
@@ -388,6 +466,101 @@ def test_vmapvjp(self, device, dtype, op):
...
+        xfail('matrix_exp'),
+        xfail('lu'),
+        xfail('fill_'),
+        xfail('block_diag'),
+        xfail('index_add'),
+        xfail('index_copy'),
+        xfail('index_put'),
+        xfail('index_fill'),
+        xfail('masked_fill'),
+        xfail('masked_scatter'),
```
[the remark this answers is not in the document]
soulitzer: Yes except for `block_diag` and `diag_embed`.

test/test_ops.py
soulitzer: This and the comment below, still don't see on the pytorch core side. Need to investigate further.

test/test_ops.py
soulitzer: Yes that would make sense. I've updated some faulty testing code as well as this list https://github.com/pytorch/pytorch/issues/66357 to reduced the discrepancy between the two. The remaining ones are:
- `Batching rule not implemented for aten::logical_{and,or}.out`
- `Batching rule not implemented for aten::_local_scalar_dense`
- `block_diag` and `diag_embed`

test/test_ops.py
zou3519: Do you have a minimal repro of this for functorch that does not use the functorch test suite? If we have a minimal repro of this for a function (say, linalg.inv) using vmap and jvp we can easily convert it to a PyTorch core test case and then use that to determine if there is actually a problem in PyTorch core or not

soulitzer: Oh wait I think I found the issue, see the latest update.

test/test_ops.py
soulitzer: Hmm it passes now for me as well

test/test_ops.py
soulitzer: These were due to me testing batched forward in addition to batched grads... but it is fixed now

test/test_ops.py
soulitzer: Update: `block_diag` should be the last one. The others have been resolved!

test/test_ops.py
```diff
@@ -385,6 +460,88 @@ def test_vmapvjp(self, device, dtype, op):
...
+        xfail('std_mean'),
+        xfail('linalg.eigvalsh'),
+
+        # These are issues that should be fixed in core. See repro in core:
+        # https://github.com/pytorch/functorch/pull/232#discussion_r751405155
+        # RuntimeError: expected scalar type double but found float
+        xfail('minimum'),
+        xfail('min', 'binary'),
+        xfail('maximum'),
+        xfail('max', 'binary'),
```
zou3519, line 517: Let's move these above the "The above failures also fail in core." line. The thing we're interested in testing is if a user uses PyTorch core and encounters the problem, not if the PyTorch test suite says there is a problem. These two things would ideally be the same but they aren't because the PyTorch test suite currently only runs jvp tests in double precision

functorch/csrc/BatchedTensorImpl.cpp
```diff
@@ -71,6 +72,7 @@ void BatchedTensorImpl::refreshSizesAndStrides() {
     sizes_and_strides_.size_at_unchecked(dim) = value_sizes.at(actual_dim);
     sizes_and_strides_.stride_at_unchecked(dim) = value_strides.at(actual_dim);
   }
+  storage_offset_= value_.storage_offset();
```
zou3519, line 75: We should rename refreshSizesAndStrides() to refreshTensorMetadata() or something

test/common_utils.py
```diff
@@ -35,7 +36,7 @@ def loop(op, in_dims, out_dim, batch_size, *batched_args, **kwarg_values):
     return loop_out
 
 
-def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3):
+def get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3, first_dim_only=False):
```
zou3519, line 39: nit: this reads better as `get_exhaustive_batched_inputs(arg_values, kwarg_values, batch_size=3, bdims=(0,  -1))`. This function however only supports `bdims=(0, -1)` and `bdims=(0,)`. If you agree that explicitly writing out the bdims is better for readability then we want to assert that bdims is either `(0,)` or `(0, -1)` to avoid needing to rewrite the function now

test/common_utils.py
```diff
@@ -72,10 +73,10 @@ def add_batch_choices(a):
             yield pytree.tree_unflatten(batched_args, arg_spec), pytree.tree_unflatten(in_dims, arg_spec), kwarg_values
 
 
-def get_fallback_and_vmap_exhaustive(op, arg_values, kwarg_values, compute_loop_out=True):
+def get_fallback_and_vmap_exhaustive(op, arg_values, kwarg_values, compute_loop_out=True, first_dim_only=False):
```
zou3519, line 76: ditto here: this reads better as `get_fallback_and_vmap_exhaustive(op, arg_values, kwarg_values, bdims=(0, 1))`

test/test_ops.py
```diff
@@ -60,23 +61,29 @@ def _autograd_grad(outputs, inputs, grad_outputs=None, retain_graph=False, creat
     return tree_unflatten(result, inputs_spec)
 
 
-def diff_arg(arg):
+def diff_arg(arg, is_jvp=False):
```
zou3519, line 64: nit: flags are hard to read. if `is_jvp=True` then we know we're dealing with jvp, if `is_jvp=False` then what APIs are we testing?

This is probably better as
```
def is_floating_point(tensor):
  ...

def tensor_requires_grad(tensor):
  return tensor.requires_grad

def diff_arg(arg, is_differentiable_fn=tensor_requires_grad):
  ...
```

test/test_ops.py
```diff
@@ -193,6 +226,48 @@ def wrapped_fn(*args, **kwargs):
...
+            self.skipTest("Skipped! NYI: inplace-testing not supported.")
+            return
+
+        for sample in samples:
+            fn, primals = normalize_op_input_output(op, sample, is_jvp=True)
+            tangents = tree_map(lambda x: torch.randn_like(x), primals)
+            primal_outs, tangent_outs = jvp(fn, primals, tangents)
+            expected_primal_outs, expected_tangent_outs = ref_jvp(fn, primals, tangents)
+            self.assertEqual(_as_tuple(primal_outs), expected_primal_outs)
+            self.assertEqual(_as_tuple(tangent_outs), expected_tangent_outs)
```
zou3519, line 269: This check is a little dangerous. If primals_out is a single Tensor, but functorch.jvp is wrong and always returns Tuples of Tensors instead of a single Tensor, we would not catch it!

zou3519 (approved):
@soulitzer could you rebase this please?

This PR LGTM. I left some comments, please read.

I'm also not sure if you were planning on resolving any of the TODOs or if you're done with this, I know that from our discussions that you weren't planning on resolving most of them.

test/test_ops.py
soulitzer: Yeah `is_jvp` is probably too jvp-centric, but I still think this argument is more of a binary thing (what would be a third function that one would pass here?). Perhaps a more "neutral" flag would be better. I'm going to rename to `requires_grad` instead for now.

test/test_ops.py
soulitzer: Good catch, I guess I just needed to update the reference implementation to not return a tuple when the size of the tuple is 1.

test/test_ops.py
zou3519: A more neutral flag sounds good to me

test/test_ops.py
soulitzer: Good point.

soulitzer:
Rebased! 

Yeah regarding the remaining TODOs, I don't think I will try to address them in this PR.

zou3519:
Closing-and-reopening to see if CI will come back...

zou3519 (approved):
LGTM as long as the tests pass

zou3519:
Failures are almost all unrelated, in the interest of developer velocity I'll add the one expected failure that the tests are reporting on. Thanks Jeffrey!

PR #234, opened 2021-11-02
add jacrev docstring

samdow:
Adds docstring for jacrev for #224, also does some minor cleanup of the readme in the `jacrev` section and removes a broken example

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,13 +169,15 @@ When composed with `vmap`, `grad` can be used to compute per-sample-gradients:
 
 ### vjp and jacrev
 
+The `vjp` transform applies `func` to `inputs` and returns a new function that
+computes vjps given some `cotangents` Tensors.
 ```
 >>> from functorch import vjp
 >>> outputs, vjp_fn = vjp(func, inputs); vjps = vjp_fn(*cotangents)
 ```
-The `vjp` transform applies `func` to `inputs` and returns a new function that
-computes vjps given some `cotangents` Tensors.
 
+The `jacrev` transform returns a new function that takes in `x` and returns the
+Jacobian of `torch.sin` with respect to `x`
 ```py
 >>> from functorch import jacrev
 >>> x = torch.randn(5)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -192,15 +194,6 @@ batched jacobians:
...
 ```
 
-`jacrev` can be composed with itself to produce hessians:
-```py
->>> def f(x):
->>>   return x.sin().sum()
->>>
->>> x = torch.randn(5)
->>> hessian = jacrev(jacrev(f))(x)
-```
```
samdow, line 202: I like this example but it's currently broken, erroring that reshape isn't supported (through a fallback or as a prim). We can get it to work with just sum, but then the hessian is all 0s and that's not super interesting

functorch/_src/eager_transforms.py
```diff
@@ -173,7 +174,73 @@ def wrapper(cotangents, retain_graph=True, create_graph=None):
...
+        >>>   return x + y ** 2
+        >>>
+        >>> x, y = torch.randn(5), torch.randn(5)
+        >>> jacobian = jacrev(f, argnums=(0,1))(x, y)
+        >>> expectedX = torch.diag(torch.ones_like(x))
+        >>> expectedY = torch.diag(2 * y)
+        >>> assert torch.allclose(jacobian[0], expectedX)
+        >>> assert torch.allclose(jacobian[1], expectedY)
+
+    .. note:
```
samdow, line 229: Since the function calls vmap and grad, figured this note also applied to `jacrev`

README.md
```diff
@@ -169,13 +169,15 @@ When composed with `vmap`, `grad` can be used to compute per-sample-gradients:
...
+The `vjp` transform applies `func` to `inputs` and returns a new function that
+computes vjps given some `cotangents` Tensors.
 ```
 >>> from functorch import vjp
 >>> outputs, vjp_fn = vjp(func, inputs); vjps = vjp_fn(*cotangents)
 ```
-The `vjp` transform applies `func` to `inputs` and returns a new function that
-computes vjps given some `cotangents` Tensors.
 
+The `jacrev` transform returns a new function that takes in `x` and returns the
```
samdow, line 179: this just fixes it so all descriptions of what's about to happen come before the code example

README.md
Chillee: wait, reshape isn't supported? That seems like something we should fix :P I'll take a look.

samdow: Oh wait this might have been an old version--I pulled main and it looks like it's working okay. Sorry about the false alarm

Issue #240, opened 2021-11-04
Actually actionable list of batching rules to write


zou3519:
For each of the items here, we should make sure all compositions have a batching rule. All of these items should be actionable (in that it is possible to write a batching rule and we are not blocked on functionalization, which is coming soon).

Note: you may need to write an OpInfo for the operator if it doesn't exist already or wait for one to be added. A lot of folks are adding OpInfos right now, so if the OpInfo doesn't exist please ask first to see if someone is working on it.

Parcel 1: top nn.functional.* and top torch.* foo
- [ ] torch.nn.functional.interpolate (this involves adding batching rules for adaptive_avg_pool1d, but we might as well do `adaptive_avg_pool{1, 2, 3}d` as well as their backward variants while we're at it)
- [ ] nn.functional.pad. The batching rule seems to have a bug; the goal is to debug it.
- [ ] nn.functional.unfold. Involves writing a batching rule for im2col, im2col_backward. I don't know how difficult these are.
- [ ] nn.functional.grid_sample. Involves writing a batching rule for the backward operator.
- [ ] torch.pow. The backward needs batching rule for logical_and ; use this as an opportunity to write the batching rules for `logical_{and, or, xor}` if those don't exist yet.

vfdev-5:
@zou3519 `nn.functional.pad` with circular option requires to do a copy: `out[..., out_d0:out_d1] = input[..., in_d0:in_d1]` and thus there is the following error is raised:
```
>           out[..., out_d0:out_d1] = input[..., in_d0:in_d1]                                                                                                                               
E           RuntimeError: vmap: aten::copy_(self, *extra_args) is not possible because there exists a Tensor `other` in extra_args that has more elements than `self`. This happened due to `other` being vmapped over but `self` not being vmapped over at level 2. Please try to use out-of-place operators instead of aten::copy_. If said operator is being called inside the PyTorch framework, please file a bug report instead. 
```
Can we do something with that ?

zou3519:
Aha. Nope, we can't do anything about that until functionalization is in, good catch.

Padarn:
Hi @zou3519, the "forward pass only" ops above means that the vjp and related operators require the functionalization too?

zou3519:
> Hi @zou3519, the "forward pass only" ops above means that the vjp and related operators require the functionalization too?

Yes, "forward pass only" means we should only try to get the `vmap` tests passing and none of the vjp/grad/compositions of {vjp, grad} tests.

[...]

PR #249, opened 2021-11-08
Fixed nn.functional.pad constant mode

vfdev-5:
Description:
- Fixed nn.functional.pad constant mode
- Updated tests

The issue was with the fact that `UNARY_POINTWISE` does not properly handle bdim > 0. The fix is to use a macro which explicitly calls `moveBatchDimToFront`.

Related to #240


vfdev-5:
@zou3519 please review this PR as well. Thanks

Chillee:
Sorry, I don't understand why UNARY_POINTWISE doesn't support bdim > 0 🤔 

Oh, is it specifically that for `constant_pad_nd` it screws up the padding argument?

Chillee (approved):

zou3519 (approved):

zou3519:
> Sorry, I don't understand why UNARY_POINTWISE doesn't support bdim > 0 🤔
> 
> Oh, is it specifically that for `constant_pad_nd` it screws up the padding argument?

yes. The padding argument is a list of pads for the last dimensions. For example, `constant_pad_nd(x: Tensor[2, 3], pad=[2, 3, 4, 5])` would pad dimension 0 using (2, 3) and dimension 1 using (4, 5).

Now, if your batch dimension is in the middle (e.g. `constant_pad_nd(x: Tensor[2, B, 3], pad=[2, 3, 4, 5])`) then we need to first move it to the front so it doesn't get padded by mistake

PR #258, opened 2021-11-12
[AOT Autograd] refactor partitioner to extract arbitrary graph given inputs and outputs

Chillee:
I refactored the graph partitioner to extract a subgraph with given inputs and outputs. In particular, it allows for inputs that aren't one of the graph placeholders.

So, for example, to save the mask, we simply find the dropout nodes, append them to the list of "saved nodes", and then call

```
fw_graph = _extract_graph_with_inputs_outputs(joint_module.graph, primal_nodes, fwd_outputs + saved_values)
bw_graph = _extract_graph_with_inputs_outputs(joint_module.graph, saved_values + tangent_nodes, bwd_outputs)
```

### DIFF
--- a/functorch/_src/eager_compilation.py
+++ b/functorch/_src/eager_compilation.py
@@ -169,13 +169,14 @@ def partition_with_recompute_fwd_in_bwd(joint_module: fx.GraphModule, _joint_inp
     for node in joint_module.graph.nodes:
         if node.target == operator.getitem and node.args[0] in dropout_nodes and node.args[1] == 1:
             saved_values.append(node)
+    primal_inputs = list(filter(is_primal, joint_module.graph.nodes))
+    tangent_inputs = list(filter(is_tangent, joint_module.graph.nodes))
     # Construct the forward module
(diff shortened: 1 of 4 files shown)


anijain2305 (approved):

byronyi:
@Chillee We seems to have a regression here. A repro:

```
aot_module(torch.nn.Dropout(), print_compile)(torch.rand((3,4),device='cuda').requires_grad_())
```

Gives forward and backward graph:

```
def forward(self, primals_1):
    native_dropout = torch.ops.aten.native_dropout(primals_1, 0.5, True)
    getitem = native_dropout[0];  native_dropout = None
    return [getitem, primals_1]

def forward(self, primals_1, tangents_1):
    native_dropout = torch.ops.aten.native_dropout(primals_1, 0.5, True);  primals_1 = None
    getitem_1 = native_dropout[1];  native_dropout = None
    native_dropout_backward = torch.ops.aten.native_dropout_backward(tangents_1, getitem_1, 2.0);  tangents_1 = getitem_1 = None
    return [native_dropout_backward]
```

byronyi:
A timely fix c910bb66d861031cd4dd6f8a2e6f6d496827c8cc resolved the issue above.

PR #259, opened 2021-11-12
Skip testing empty_like instead of xfail

zou3519:
This test is flaky.


Chillee:
I think we need this skip elsewhere as well.

PR #263, opened 2021-11-15
Added adaptive_max_poolNd batch rule

vfdev-5:
Description:
- Added adaptive_max_poolNd fw/bw batch rules
- Updated tests

Related to #240

Notes:
I created two additional macros to handle adaptive_max_pool2d and adaptive_max_pool3d_backward.
Not sure if we could make a generic rule to handle max_pool2d_with_indices_backward_batch_rule and adaptive_max_pool3d_backward, as max_pool2d_with_indices_backward_batch_rule requires some args in the middle between gradOutput, input and indices.


Chillee:
@vfdev-5 Could you use one of the boxed fallbacks to handle this? Or modify one of the boxed fallbacks?

vfdev-5:
@Chillee which boxed fallbacks you think of ?

Chillee:
Looks similar at first glance to `EXISTING_DIM_ALL_BOXED`.

vfdev-5:
@Chillee I updated the PR according to your suggestion

functorch/csrc/BatchRulesPooling.cpp
```diff
@@ -82,6 +82,46 @@ max_pool2d_with_indices_batch_rule(
       reshape_dim_outof(0, bdim_size, std::get<1>(result)), 0);
 }
 
+// We can't use ALL_TENSORS_HAVE_OPTIONAL_BDIM_BOXED because the CUDA
```
Chillee, line 85: Perhaps it's better to just augment the boxed rule with the option to make some args contiguous?

vfdev-5: @Chillee good idea! I did an implementation with one tensor that could be contiguous (later we can make it more generic if needed). What do you think ?

Chillee: Looks good to me!

PR #271, opened 2021-11-16
strict=True for grad, vjp

zou3519:
Test Plan:
- run tests


zou3519:
Actually I'm not sure if this should be a thing -- need to check how well it composes with vmap

PR #281, opened 2021-11-19
Added decomposition testing + display infra

Chillee:
https://docs.google.com/spreadsheets/d/1Sp4HUjxwMifS5oDQg0yvjqk7hKOpCfKO4jWH4MTGP-k/edit#gid=1815506345

This tracks the decompositions written.


functorch/_src/python_key.py
```diff
@@ -27,33 +29,73 @@ def decomposition_decorator(f):
...
     return out_grad * (y * (1 - y))
 
-@register_decomposition(aten._s_where)
-def _s_where_decomposition(a, b, c):
-    return aten.where(a, b, c)
-
 @register_decomposition(aten.detach)
-def detach_decomposition(x):
+# This is valid if we always sit behind autograd
+def detach_decomposition(x: Tensor):
```
zou3519, line 47: Actually I'm confused: why do you see detach() in `__torch_dispatch__` if we sit underneath autograd?

functorch/_src/python_key.py
```diff
@@ -27,33 +29,73 @@ def decomposition_decorator(f):
...
     return x
 
 @register_decomposition(aten.softplus_backward)
-def softplus_backward_decomposition(out_grad, x, beta, threshold, out):
+# The out argument seems to always be ignored?
+def softplus_backward_decomposition(out_grad: Tensor, x: Tensor, beta: float, threshold: float, out):
```
zou3519, line 52: Goes into the "things we should fix" decompositions

test/test_ops.py
```diff
@@ -598,8 +600,160 @@ def test_vjpvmap(self, device, dtype, op):
 
                 self.assertEqual(result_vjps, expected_vjps)
 
+class InplaceError(Exception):
+    def __repr__(self):
+        return "Decomposition Tensor with no elem was created (probably due to an in-place op)"
```
zou3519, line 605: Do you have a sense of how this happens? Maybe it's a torch_dispatch bug

test/test_ops.py
```diff
@@ -598,8 +600,160 @@ def test_vjpvmap(self, device, dtype, op):
...
+        skip('nanmean'),
+    })
+    def test_decomposition(self, device, dtype, op):
+        if dtype not in op.supported_dtypes(dtype):
+            self.skipTest("Dtype not in op's supported dtypes")
+            return
+        if is_inplace(op, op.get_op()):
+            self.skipTest("op is inplace")
+            return
+        print(op.op)
```
zou3519, line 683: stray print

test/test_ops.py
```diff
@@ -598,8 +600,160 @@ def test_vjpvmap(self, device, dtype, op):
...
+
+        except InplaceError:
+            self.skipTest("op is inplace")
+            return
+        except RuntimeError as e:
+            if "not implemented for 'Half'" in str(e):
+                self.skipTest("(backwards) op not supported for Half")
+                return
+            if "Mismatch in shape: grad_output" in str(e):
+                self.skipTest("Some weird issue with autograd engine and tensor subclasses")
```
zou3519, line 743: Is the problem that you need sum_to in the decomposition rules? Or that the tensor subclass isn't propagating the shape properly?

test/test_ops.py
```diff
@@ -598,8 +600,160 @@ def test_vjpvmap(self, device, dtype, op):
...
+            raise e
+
+    def test_placeholder(self):
+        with open('op_analysis/run_ops.txt', 'w') as f:
+            for op in DecompositionTensor.run_ops:
+                f.write(f'{op.__name__}\n')
+        with open('op_analysis/run_decompositions.txt', 'w') as f:
+            for op in DecompositionTensor.run_decompositions:
+                f.write(f'{op.__name__}\n')
+
```
zou3519, line 753: What's this for?

test/test_ops.py
```diff
@@ -598,8 +600,160 @@ def test_vjpvmap(self, device, dtype, op):
...
+                else:
+                    args = [sample_input.input] + list(sample_input.args)
+                    kwargs = sample_input.kwargs
+                    orig_out = func(*args, **kwargs)
+
+                    args = tree_map(wrap_tensor, args)
+                    kwargs = tree_map(wrap_tensor, kwargs)
+                    decomp_out = func(*args, **kwargs)
+                    _assertEqual(orig_out, tree_map(unwrap_tensor, decomp_out))
```
zou3519, line 732: cool!

test/test_ops.py
Chillee: I keep track of which ops are run in the OpInfo tests and which decompositions are run so that I can dump them out to a file, which gets loaded here: https://docs.google.com/spreadsheets/d/1Sp4HUjxwMifS5oDQg0yvjqk7hKOpCfKO4jWH4MTGP-k/edit#gid=1815506345

test/test_ops.py
Chillee: I'm... not sure. Need to investigate a bit more.

functorch/_src/python_key.py
Chillee: I think the autograd engine calls detach. And it's not a composite op.

test/test_ops.py
Chillee: I don't think so - it's something to do with the tensor subclass, probably in combination with in-place ops.

For example, `__rmatmul__` fails this way and does'nt hit any decompositions.

PR #283, opened 2021-11-22
Layer norm decomposition

zou3519:


zou3519:
cc @Chillee -- if you want a decomposition for layer norm, this PR is the decomposition of the forward and https://gist.github.com/zou3519/af1cc49a2854f66f7467944b0a861209 is a decomposition of the backward.

It turns out we can write faster batching rules without decomposition so I'm going to close this

PR #306, opened 2021-12-02
Added cdist forward/backward batching rules

vfdev-5:
Description:
- Added cdist forward/backward batching rules
- updated tests
  - added `a.select(in_dim, idx).contiguous()` when computing `loop` expected values as required by cdist_backward to have contiguous input.

Related to #240


functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -285,6 +380,10 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   REDUCTION_BOXED(any.dim);
   REDUCTION_BOXED(argmax);
   REDUCTION_BOXED(argmin);
+  VMAP_SUPPORT("cdist", cdist_batch_rule);
+  VMAP_SUPPORT("_euclidean_dist", euclidean_dist_batch_rule);
```
vfdev-5, line 384: Maybe, there is a similar macro to apply `moveBatchDimToFront` on all input tensors ...

functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -275,6 +275,101 @@ std::tuple<Tensor, optional<int64_t>, Tensor, optional<int64_t>> aminmax_batchin
...
+  auto x12 = cdist_batch_rule_helper(x1_, x1_bdim, x2, x2_bdim);
+  x1_ = std::get<0>(x12);
+  Tensor x2_ = std::get<1>(x12);
+
+  auto grad_ = moveBatchDimToFront(grad, grad_bdim);
+  if ((x1_bdim || x2_bdim) && !grad_bdim) {
+    auto bs = get_bdim_size2(x1_, 0, x2_, 0);
+    grad_ = ensure_has_bdim(grad_, grad_bdim.has_value(), bs);
+    grad_ = grad_.contiguous();
+  }
```
vfdev-5, line 361: It was very tricky to figure out what to do with batched/non-batched tensors x1, x2, grad etc. Maybe, something could be simplified...

functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -285,6 +380,10 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   REDUCTION_BOXED(any.dim);
   REDUCTION_BOXED(argmax);
   REDUCTION_BOXED(argmin);
+  VMAP_SUPPORT("cdist", cdist_batch_rule);
```
zou3519, line 383: cdist is a composite operation, right? Why didn't we just do an `OP_DECOMPOSE` for it?

functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -275,6 +275,101 @@ std::tuple<Tensor, optional<int64_t>, Tensor, optional<int64_t>> aminmax_batchin
...
+  while (x2_bdim && x1_.dim() - d1 > x2_.dim() - d2) {
+    // x1:                      [  S1, S2, .............., SK]
+    // x2: [B, D1, ..., DN]  -> [B, 1,  1, ..., 1, D1 ..., DN]
+    x2_ = x2_.unsqueeze(1);
+  }
+  while (x1_bdim && x2_.dim() - d2 > x1_.dim() - d1) {
+    // x2:                      [  S1, S2, .............., SK]
+    // x1: [B, D1, ..., DN]  -> [B, 1,  1, ..., 1, D1 ..., DN]
+    x1_ = x1_.unsqueeze(1);
+  }
```
zou3519, line 296: I'm confused -- what is going on here?

functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -275,6 +275,101 @@ std::tuple<Tensor, optional<int64_t>, Tensor, optional<int64_t>> aminmax_batchin
...
+  auto x12 = cdist_batch_rule_helper(x1, x1_bdim, x2, x2_bdim);
+  return std::make_tuple(at::cdist(std::get<0>(x12), std::get<1>(x12), p, compute_mode), 0);
+}
+
+std::tuple<Tensor,optional<int64_t>> cdist_forward_batch_rule(
+    const Tensor& x1, optional<int64_t> x1_bdim,
+    const Tensor& x2, optional<int64_t> x2_bdim,
+    const double p, c10::optional<int64_t> compute_mode) {
+
+  auto x12 = cdist_batch_rule_helper(x1, x1_bdim, x2, x2_bdim);
```
zou3519, line 321: cdist_batch_rule_helper returns tensors that are more than 3 dimensional. Then we call `_cdist_forward` in the following line.

In the [PyTorch documentation for cdist](https://pytorch.org/docs/stable/generated/torch.cdist.html) it seems to imply that cdist accepts tensors with 3 dimensions only. Is the PyTorch documentation wrong?

zou3519: Reading the OpInfo for this, it does look like the pytorch documentation is wrong unless I am reading the incorrect page

functorch/csrc/BatchRulesReduceOps.cpp
```diff
@@ -275,6 +275,101 @@ std::tuple<Tensor, optional<int64_t>, Tensor, optional<int64_t>> aminmax_batchin
...
+  return std::make_tuple(at::cdist(std::get<0>(x12), std::get<1>(x12), p, compute_mode), 0);
+}
+
+std::tuple<Tensor,optional<int64_t>> cdist_forward_batch_rule(
+    const Tensor& x1, optional<int64_t> x1_bdim,
+    const Tensor& x2, optional<int64_t> x2_bdim,
+    const double p, c10::optional<int64_t> compute_mode) {
+
```
zou3519, line 320: We should probably validate that x1 and x2 have logical dimension >= 2. Otherwise we might throw a weird error in the batching rule

functorch/csrc/BatchRulesReduceOps.cpp
zou3519: Is this the same logic as https://github.com/pytorch/functorch/blob/b1d08d4669596a83e1d4e3fa72842bc224dff531/functorch/csrc/BatchRulesBinaryOps.cpp#L48-L53 ?

functorch/csrc/BatchRulesReduceOps.cpp
vfdev-5: Let me check that, I haven't thought about that. Thanks for the suggestion.

functorch/csrc/BatchRulesReduceOps.cpp
vfdev-5: Yes, the docs is incomplete about what could be accepted as input. I think it should be something like
- x1: *, P, M
- x2: *, R, M

vfdev-5:
@zou3519 thanks for the review, I'll update the PR accordingly

zou3519:
@vfdev-5 No rush, I'm still making my way through it (need to read the backward batch rule)

functorch/csrc/BatchRulesReduceOps.cpp
vfdev-5: Looks like `_binary_pointwise_batch_rule` + making contiguous `tensor_` and `other_` can replace all the code for `_cdist_forward`. Thanks a lot for pointing out `maybePadToLogicalRank`.

functorch/csrc/BatchRulesReduceOps.cpp
vfdev-5: By default, right now the error looks like that (if passing 2d tensor before batching):
```
RuntimeError: cdist only supports at least 2D tensors, X1 got: 1D
```
I think this is readable. What do you think ?

Code:
```python
def func(x1, x2):
    return torch.cdist(x1, x2, 1.0, 'use_mm_for_euclid_dist')

t1 = torch.rand(4, 5)
t2 = torch.rand(3, 5)

out = vmap(func, in_dims=(0, None))(t1, t2)
```

zou3519: Thank you for checking. Yes, that is readable -- let's leave the code as-is

vfdev-5:
@zou3519 I moved the code to BatchRulesBinaryOps.cpp to take benefit of `BINARY_POINTWISE` that I had to modify a little bit. I rewrote as well the backward pass and left comments about why I had to modify the inputs that way. Could you please review and let me know if we can simplify it more. Thanks

test/common_utils.py
```diff
@@ -25,7 +25,9 @@ def loop(op, in_dims, out_dim, batch_size, *batched_args, **kwarg_values):
         flat_args, args_spec = pytree.tree_flatten(batched_args)
         flat_dims, dims_spec = pytree.tree_flatten(in_dims)
         assert(args_spec == dims_spec)
-        new_args = [a.select(in_dim, idx) if in_dim is not None else a for a, in_dim in zip(flat_args, flat_dims)]
+        # Tensors in new_args could be required to be contiguous for certain ops like cdist and cdist_backward
+        # Thus, a.select(...) -> a.select(...).contiguous()
+        new_args = [a.select(in_dim, idx).contiguous() if in_dim is not None else a for a, in_dim in zip(flat_args, flat_dims)]
```
zou3519, line 30: Do we actually need this change?
- If cdist does not support non-contiguous inputs, then that's a bug.
- cdist_backward should not get directly invoked by `loop` because it only exists in aten.

functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -218,6 +297,10 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   UNARY_POINTWISE(clamp_max);
   POINTWISE_BOXED(clamp_max_);
 
+  VARIADIC_BDIMS_BOXED(_euclidean_dist);
+  BINARY_POINTWISE_CONTIG(_cdist_forward);
```
zou3519, line 301: Why do we need to make the inputs for this contiguous?

If the backward pass requires saved inputs to be contiguous, then we should handle that in the backward batch rule (.contiguous() is generally an expensive operation and if a user never calls .backward() then we don't want them to pay for it)

functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -163,11 +194,59 @@ Tensor addr_decomposition(
...
+  auto x1_ = x1;
+  if (cdist_bdim && !x1_bdim) {
+    // We need to make sure that x1 has batch dim if cdist has one
+    // otherwise, we get
+    // RuntimeError: Function CdistBackward0 returned an invalid gradient at index 1 - got [5]
+    // but expected shape compatible with [4, 5]
+    auto bs = cdist.size(*cdist_bdim);
+    x1_ = ensure_has_bdim(x1, false, bs);
+    x1_ = x1_.contiguous();
+    x1_bdim = 0;
```
zou3519, line 213: Hmm, what happens if you don't create the batch dim on x1? I don't know if you've tried this already, but the error message suggests another potential solution: if the output of cdist_backward_batch_rule is a regular Tensor with no batch dim, then we should return `std::make_tuple(regular_tensor, nullopt)`

functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -163,11 +194,59 @@ Tensor addr_decomposition(
...
+  // _binary_pointwise_batch_rule
+  auto x12 = _binary_pointwise_helper<true>(x1_, x1_bdim, x2, x2_bdim);
+  x1_ = std::get<0>(x12);
+  auto x2_ = std::get<1>(x12);
+
+  auto grad_ = moveBatchDimToFront(grad, grad_bdim);
+  if ((x1_bdim || x2_bdim) && !grad_bdim) {
+    // We need to make sure that grad has batch dim if x1 or x2 have one
+    // Probably, there is an assumption on the strides.
+    // Otherwise grad input contains thrash values, e.g. -7.0816e+29, 7.0816e+29
```
zou3519, line 226: This seems reasonable, in the non-batched case if x1 has a dimension at the front that x2 doesn't have, then grad must have it as well

functorch/csrc/BatchRulesBinaryOps.cpp
zou3519: Reading the rest of the code it looks like you may have tried that already

zou3519 (commented):
This looks pretty good to me. Two questions but the logic looks correct

functorch/csrc/BatchRulesBinaryOps.cpp
vfdev-5: Original cdist op is working such that forward pass does not fail on non-contig input but backward pass fails as requires contiguous input:
```python
x = torch.rand(4, 32, 5)[:, 0, :]
y = torch.rand(6, 32, 5)[:, 0, :]
assert not x.is_contiguous()
assert not y.is_contiguous()

x.requires_grad_()
_ = y.requires_grad_()

ret = torch.cdist(x, y)
ret = ret.sum()
ret.backward()

> RuntimeErrorTraceback (most recent call last)
<ipython-input-44-c6bee6e0f2fc> in <module>
      9 ret = torch.cdist(x, y)
     10 ret = ret.sum()
---> 11 ret.backward()

/opt/conda/lib/python3.8/site-packages/torch/_tensor.py in backward(self, gradient, retain_graph, create_graph, inputs)
    305                 create_graph=create_graph,
    306                 inputs=inputs)
--> 307         torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
    308 
    309     def register_hook(self, hook):

/opt/conda/lib/python3.8/site-packages/torch/autograd/__init__.py in backward(tensors, grad_tensors, retain_graph, create_graph, grad_variables, inputs)
    152         retain_graph = create_graph
    153 
--> 154     Variable._execution_engine.run_backward(
    155         tensors, grad_tensors_, retain_graph, create_graph, inputs,
    156         allow_unreachable=True, accumulate_grad=True)  # allow_unreachable flag

RuntimeError: _cdist_backward requires X1 to be contiguous
```

Maybe, original backward pass could simply call `.contiguous()` on non-contig input inside backward pass instead of failing in pytorch.

From functorch side, the ops like vjpvmap do not call backward batch rule but only forward batch rule and original backward pass and thus failing with the same as above message.

functorch/csrc/BatchRulesBinaryOps.cpp
vfdev-5: Yes, maybe we can simplify that somehow...

test/common_utils.py
vfdev-5: Without this patch, expected results can't be computed for certain batched inputs due to the same issue as here: https://github.com/pytorch/functorch/pull/306/files/04ae2b1dbd3140b389a684c75c1cc945f24e2180#r769666360

cdist supports non-contig inputs but for forward pass only.

functorch/csrc/BatchRulesBinaryOps.cpp
zou3519: Oh yeah, this is a PyTorch bug

zou3519: https://github.com/pytorch/pytorch/issues/69997 --> We should have cdist_backward call .contiguous in PyTorch. In general, if an operation is supported in the forward pass by PyTorch, then the backward pass should also be supported.

functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -163,6 +175,52 @@ Tensor addr_decomposition(
...
+    // but expected shape compatible with [4, 5]
+    auto bs = cdist.size(*cdist_bdim);
+    x1_ = ensure_has_bdim(x1, false, bs);
+    x1_ = x1_.contiguous();
+    x1_bdim = 0;
+  }
+
+  // We need to apply the same preprocessing on x1 and x2 as in the forward pass
+  // _binary_pointwise_batch_rule
+  auto x12 = _binary_pointwise_helper(x1_, x1_bdim, x2, x2_bdim);
```
zou3519, line 199: Does cdist do type promotion? `_binary_pointwise_helper` handles type promotion; if cdist doesn't do type promotion then this will lead to different behavior than cdist

functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -218,6 +276,10 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   UNARY_POINTWISE(clamp_max);
   POINTWISE_BOXED(clamp_max_);
 
+  VARIADIC_BDIMS_BOXED(_euclidean_dist);
+  BINARY_POINTWISE(_cdist_forward);
```
zou3519, line 280: It looks like cdist requires the dtypes of the input tensors to be the same. The batching rule should not be more permissive than the original operator, so we shouldn't use BINARY_POINTWISE here; we probably want a variant of BINARY_POINTWISE that does not do type promotion

image

zou3519: comparison_pointwise_batch_rule seems to not do type promotion, maybe we could reuse that or do some refactoring

zou3519 (commented):
thanks for fixing the contiguity problem in pytorch/pytorch! I had one last comment on if cdist has type promotion

functorch/csrc/BatchRulesBinaryOps.cpp
vfdev-5: Thanks for a thourough review @zou3519 ! Yes, _binary_pointwise_helper performs a dtype promotion if one arg is a scalar, but cdist can't work with scalar, at least 2d tensors: https://github.com/pytorch/pytorch/blob/8a912014b166c97fcaf2b89e6ff5a466d1a93f7b/aten/src/ATen/native/Distance.cpp#L119-L120

So, using `_binary_pointwise_helper` on scalars we change the type but cdist will complain about dimensions.
If we are using correct shapes and incorrect dtypes, IMO there is no dtype promotion by cdist batch rule and original cdist will complain about non-floating dtype...

functorch/csrc/BatchRulesBinaryOps.cpp
vfdev-5: The error message we see on the image is raised by `matmul` in https://github.com/pytorch/pytorch/blob/8a912014b166c97fcaf2b89e6ff5a466d1a93f7b/aten/src/ATen/native/Distance.cpp#L44 but it should be caught a bit earlier for the second arg (there is a typo in the code, see https://github.com/pytorch/pytorch/pull/70178)

functorch/csrc/BatchRulesBinaryOps.cpp
zou3519: Good point, thanks for the reasoning!

zou3519 (approved):
lgtm!

functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -39,9 +37,11 @@ std::tuple<Tensor,optional<int64_t>> _binary_pointwise_batch_rule(
   auto tensor_is_logical_scalar = (tensor_logical_rank == 0 && tensor_batch_dim.has_value());
   auto other_is_logical_scalar = (other_logical_rank == 0 && other_batch_dim.has_value());
   if (tensor_is_logical_scalar && !other_is_logical_scalar) {
+    std::cout << "HERE 1" << std::endl;
     handleScalarTypePromotion(tensor_, other_);
   }
   if (other_is_logical_scalar && !tensor_is_logical_scalar) {
+    std::cout << "HERE 2" << std::endl;
```
zou3519, line 44: You probably want to delete this before I merge this?

vfdev-5: Sorry about that :) yes, this shouldn't be there

PR #309, opened 2021-12-03
[Recompute] Fixes observed during dropout profiling

anijain2305:
@Chillee


Chillee:
Fixes already merged in a separate commit.

PR #315, opened 2021-12-06
[batching] diagonal_scatter

kshitij12345:


functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -425,13 +425,41 @@ Tensor select_scatter_decomp(
...
+  // we need `result` to be contiguous as we are going to write `src`
+  // With expanded `self_` tensor we may end up overwriting values.
+  auto result = self_.contiguous();
+
+  auto self_logical_rank = rankWithoutBatchDim(self, self_bdim);
+  dim1 = maybe_wrap_dim(dim1, self_logical_rank) + 1;
+  dim2 = maybe_wrap_dim(dim2, self_logical_rank) + 1;
+
+  // mutates `result`: updates its diagonal to be filled with `src_`.
+  at::diagonal(result, offset, dim1, dim2).copy_(src_);
```
zou3519, line 450: in-place ops are not good to have in a batching rule due to all of our problems with in-place operations

functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -425,13 +425,41 @@ Tensor select_scatter_decomp(
   return at::scatter(self, dim, index_.expand_as(self), source.unsqueeze(dim).expand_as(self));
 }
 
+std::tuple<Tensor, optional<int64_t>> diagonal_scatter_batch_rule(
+    const Tensor &self, c10::optional<int64_t> self_bdim,
+    const Tensor &src, c10::optional<int64_t> src_bdim,
+    int64_t offset, int64_t dim1, int64_t dim2)
```
zou3519, line 431: is it possible to call diagonal_scatter in the batch rule?

kshitij12345: Ah that makes sense. Thanks!

zou3519 (approved):
LGTM!

PR #319, opened 2021-12-07
Single cache

anijain2305:
Replacing with single cache. This one removes the use of templates to have a single cache, which is much easier to control. 

The key of the cache is of type string, which is built by naively concatenating all the private members of the SpecializationKey class (either Dynamic or Static for now).

Some of the missing items before merging
* Testing the latency. String building might make look up slow.


Chillee:
Some benchmark numbers after optimizing it. Time is in microseconds
```
for 2 args

main: 0.95
PR: 2.47
optimized: 0.90

for 100 args

main: doesn't work lol (but if you update it to work specifically for 100 you get 37)
PR: 90.73
updated: 11.0
```

PR #322, opened 2021-12-08
[batch-rule] householder_product

kshitij12345:
Reference: https://github.com/pytorch/functorch/issues/240


functorch/csrc/BatchRulesLinearAlgebra.cpp
```diff
@@ -160,6 +178,7 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   VMAP_SUPPORT("mv", mv_batch_rule);
   VMAP_SUPPORT("mm", mm_batch_rule);
   m.impl("linear", linear_decomp);
+  m.impl("orgqr", orgqr_decomp);
```
kshitij12345, line 181: Without this we get 
`RuntimeError: aten::orgqr hit the vmap fallback which is currently disabled`

But `orgqr` is a composite operator 🤔.

vfdev-5: Maybe, you can add it into BatchRulesDecompositions.cpp

functorch/csrc/BatchRulesLinearAlgebra.cpp
```diff
@@ -151,6 +151,20 @@ Tensor linear_decomp(
   return result;
 }
 
+std::tuple<Tensor, c10::optional<int64_t>>
```
Chillee, line 154: Can we use `EXISTING_BDIM` or `EXISTING_BDIM_ALL_BOXED` for this?

kshitij12345: Unfortunately neither of them works. 
* `EXISTING_BDIM` seems to assume that only `self` has bdim 
* With `EXISTING_BDIM_ALL_BOXED` we get `torch.linalg.householder_product: input.shape[-1] must be greater than or equal to tau.shape[-1]`

zou3519:
@Chillee @kshitij12345 can we merge this?

Chillee:
@zou3519 I'm fine with merging it.

zou3519:
Merged in https://github.com/pytorch/functorch/pull/972

Issue #305, opened 2021-12-02
Check how functorch interacts with autocast


zou3519:
I have a feeling it doesn't

vfdev-5:
Vmap seems working with autocast
```python
# Creates some tensors in default dtype (here assumed to be float32)
a_float32 = torch.rand(4, 2, 3, device="cuda")
b_float32 = torch.rand(4, 3, 2, device="cuda")
c_float32 = torch.rand(4, 2, 2, device="cuda")
d_float32 = torch.rand(4, 3, 2, device="cuda")

def func(x, y, z, w):
    with autocast():
        e_float16 = torch.matmul(x, y)
        assert e_float16.dtype == torch.float16, e_float16.dtype
        f_float16 = torch.matmul(z, e_float16)
        assert f_float16.dtype == torch.float16, f_float16.dtype
    return torch.matmul(w, f_float16.float())

expected = func(a_float32, b_float32, c_float32, d_float32)
print(expected.shape)

out = vmap(func, in_dims=0)(a_float32, b_float32, c_float32, d_float32)
print(out.shape)

assert expected.allclose(out)

@autocast()
def func(x, y, z, w):
    e_float16 = torch.matmul(x, y)
    assert e_float16.dtype == torch.float16, e_float16.dtype
    f_float16 = torch.matmul(z, e_float16)
    assert f_float16.dtype == torch.float16, f_float16.dtype
    return torch.matmul(w, f_float16)

expected = func(a_float32, b_float32, c_float32, d_float32)
print(expected.shape)

out = vmap(func, in_dims=0)(a_float32, b_float32, c_float32, d_float32)
print(out.shape)

assert expected.allclose(out)
```

vfdev-5:
@zou3519 can you detail in which case you think it wont work ?

zou3519:
Hmm, so I have three questions:

## What do we expect to be the behavior of the following?

Does the position of the context manager change anything?
```
def bar(x):
  with autocast():
    ...

# case 1: context manager inside function
vmap(bar)(x)

# case 2: context manager outside
with autocast():
  vmap(foo)(x)
```

## Do we know when autocast runs?

We would want all of the functorch transforms to run before autocast. That is, if we do `vmap(grad(func))(x, y)` where `func` has autocast, then:
1. x and y get wrapped in TensorWrapper(BatchedTensor(..))
2. We should dispatch to the autograd implementation of e.g. matmul
3. This will redispatch to the batching rule for matmul
[...]

PR #327, opened 2021-12-09
Added tests for vmap and autocast

vfdev-5:
Description:
- Added tests for vmap and autocast on CUDA/CPU
  - Note that tests are failing somehow on CPU
- Removed unused imports in test_vmap.py

Related to #305

### DIFF
--- a/test/test_vmap.py
+++ b/test/test_vmap.py
@@ -4,20 +4,19 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from unittest.case import skipIf, skip
 from torch.testing._internal.common_utils import TestCase, run_tests
 import torch
 import torch.nn.functional as F
-import torch.utils._pytree as pytree
 from torch import Tensor
 import functools
 import itertools
 import textwrap
 import warnings
 import unittest
-import re
 from torch.testing._internal.common_device_type import instantiate_device_type_tests, \
     skipCUDAIfNoMagma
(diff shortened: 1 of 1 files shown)


test/test_vmap.py
```diff
@@ -999,6 +995,66 @@ def foo(x):
...
+
+        with torch.autocast(dtype=amp_dtype, device_type=device):
+            expected = func3(a_float32, b_float32, c_float32, d_float32)
+            out = vmap(func3)(a_float32, b_float32, c_float32, d_float32)
+
+        assert expected.allclose(out)
+
+    @skip("Somehow, vmap and autocast do not work on CPU")
+    def test_vmap_autocast_cpu(self):
+        self._test_vmap_autocast("cpu")
```
zou3519, line 1052: Interesting. Do you have a stack trace?

zou3519 (approved):

test/test_vmap.py
vfdev-5: The reason why this test is failing is that matmul uses decomposes to bmm which has at::matmul in the batching rule.
But matmul with autocast on CPU only goes to fallthrough: https://github.com/pytorch/pytorch/blob/c76c6e9bd345ba697ea2a6fb8c2e24f051bb2e00/aten/src/ATen/autocast_mode.cpp#L465-L480 and autocasted on GPU : https://github.com/pytorch/pytorch/blob/c76c6e9bd345ba697ea2a6fb8c2e24f051bb2e00/aten/src/ATen/autocast_mode.cpp#L331

PR #346, opened 2021-12-15
Fixed python code formatting and added flake8 setup

### LINKED ISSUE #307: Set up some linters
Like flake8

vfdev-5:
Fixes #307


samdow:
Hi again @vfdev-5! Sorry I missed a lot of the conversation on the issue and thanks for setting this up!

One thing that I wanted to flag: `pytorch/pytorch` [uses github actions for this](https://github.com/pytorch/pytorch/blob/master/.github/workflows/lint.yml). Functorch should have access to github actions (it's just we don't have any CI systems that use it). I was wondering what your thoughts were about using github actions and eventually doing everything in `quick-checks`. If you just want to do flake8 for now, that makes sense

vfdev-5:
Hi @samdow , I have no preferences on using either circleci or GA both have pros/cons. Yes, we can also setup GA quick-checks similar to pytorch. PyTorch's one does a lot of checks and if for functorch we decide to make checks on python code formatting and C++ code formatting it would make sense to take a subset of what's done there and put it here.
In this PR I did a very simple setup and show how python formatted code looks like.
If you think that we need to move to GA quick-checks right now, I'm happy to make it instead of current setup.

samdow:
I would say that if you're able to make the flake8 a GA now, that would be the best because we'll likely have to switch in the future for some of the other checks (specifically ones that use git). If this proves to be too buggy or difficult, definitely feel free to leave this as is

vfdev-5:
> eventually doing everything in `quick-checks`

Oh, I just got your question. I think majority of these checks are caught by flake8 on python files.
Once we would like to make C++ code better formatted we can use all those checks. Maybe, we can add them in another PR such that this one do not modify every single file

samdow:
> such that this one do not modify every single file

I see! I definitely forgot how much updating there was going to be on our files because of adding a linter. Totally makes sense to wait on everything

vfdev-5:
@samdow I updated the PR and now we have to enable GHA on the repo I think...

samdow:
I'm seeing this which is extra weird...
Screen Shot 2021-12-16 at 6 21 45 PM
I think from [reading around](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks) that they don't enable running on github from forks by default. But I don't see the button from the docs in my browser. I tried adding some permissions and I'll check if they've updated in ~1 hour

Otherwise, I'll talk with Horace + Richard tomorrow morning about if it makes more sense to switch this back to circle ci or to merge blindly and fix any errors after. Thanks for your patience! I realize that more time with this standing just makes more work for you, so we'll figure out a resolution ASAP

samdow:
Okay I'm not seeing the button. Out of an abundance of caution, I think we should revert this to run on circle ci and I'll figure out the GHA story soon. So sorry about all the churn

vfdev-5:
@samdow I had a similar problem on another repository. I had to create a dummy workflow like labeler and deactivate it late and thus forked PRs with additional workflow could be executed.
On the other hand, I could run the GHA on my fork: https://github.com/vfdev-5/functorch/runs/4554385503?check_suite_focus=true

Finally, I'll push a bit more changes: right now flake8 checks only functorch folder but PR modifies examples and codegen etc. I'll fix all remaining errors with other folders and make flake8 run on them as well

test/test_vmap.py
```diff
@@ -2698,12 +2701,12 @@ def test_conv2d(self):
             for loop_out, batched_out in get_fallback_and_vmap_exhaustive(conv_fn, arg_values, kwarg_values):
                 self.assertEqual(loop_out, batched_out)
 
-            mod2 = torch.nn.Conv2d(4, 8, kernel_size=3, groups=2, stride=3, padding=1, dilation = 2)
-            arg_values = [torch.randn(inp_shape), mod.weight, mod.bias]
+            mod2 = torch.nn.Conv2d(4, 8, kernel_size=3, groups=2, stride=3, padding=1, dilation=2)
```
samdow, line 2704: Okay from the test failure, let's try having this be `conv_mod` instead of `torch.nn.Conv2d`? We can also disable this part of the test and file an issue if this is blocking

vfdev-5: Oh, I see. I replaced a bit blindly unused variable: `mod -> mod2`. Thanks for pointing out @samdow !

samdow:
Okay one final rebase after Richard's test. Mostly I haven't fully convinced myself that the conv2d test will pass even with the change because the math around strides/padding/groups/etc. doesn't seem right. If it doesn't pass, we have an OpInfo test for conv2d now so we can just remove that whole test

samdow:
I'm going to merge this since it looks like all the errors are the same as on master and flake8 looks okay

Issue #320, opened 2021-12-07
Register things in torch.return_types as PyTrees


zou3519:
## Problem

vmap over a function that returns torch.return_types ends up ignoring the torch.return_types:
```
import torch
from functorch import vmap

result = vmap(torch.max, (0, None))(torch.randn(3, 2), 0)
print(result)
# (tensor([-0.7059, -0.0751,  0.9852]), tensor([0, 0, 1]))

torch.max(torch.randn(2), 0)
# torch.return_types.max(
# values=tensor(2.0291),
# indices=tensor(1))
```

`vmap(torch.max, (0, None))(torch.randn(3, 2), 0)` should return something of type torch.return_types.

vfdev-5:
@kshitij12345 can we close this issue as #350 was merged ?

kshitij12345:
Yes. Thanks :)

PR #350, opened 2021-12-16
[pytree] register return_types

kshitij12345:
Reference: https://github.com/pytorch/functorch/issues/320

TODO:
* [x] Add tests


functorch/_src/pytree_hacks.py
```diff
@@ -7,20 +7,28 @@
...
-# We probably want some special behavior for named tuples?
 def tree_flatten_hack(pytree):
     if _pytree._is_leaf(pytree) and not isinstance(pytree, tuple):
         return [pytree], _pytree.LeafSpec()
 
-    if isinstance(pytree, tuple):
-        typ = tuple
-    else:
-        typ = type(pytree)
+    typ = type(pytree)
```
zou3519, line 31: Can we delete tree_flatten_hack now?

test/test_vmap.py
```diff
@@ -3435,6 +3435,13 @@ def op(inp, running_mean, running_var, weight, bias, training):
 
         test(self, op, tuple(inputs), in_dims=tuple(in_dims))
 
+    def test_namedtuple_returns(self, device):
```
zou3519, line 3438: nit: we're testing torch.return_types returns. Let's change the name of this to `test_torch_return_types_returns` or something

functorch/_src/pytree_hacks.py
```diff
@@ -4,38 +4,7 @@
...
-# TODO: The following function should only be used with vmap.
-# torch.return_types should be registered as PyTree nodes.
-# I can't figure out how to do that, so we are turning all of them
-# into normal Tuples for now (this is what vmap used to do anyways).
-# We probably want some special behavior for named tuples?
-
-
-def tree_flatten_hack(pytree):
-    if _pytree._is_leaf(pytree) and not isinstance(pytree, tuple):
-        return [pytree], _pytree.LeafSpec()
```
zou3519, line 20: I don't remember if this hack was for namedtuples or torch.return_types or both. Could we add a test for functions that accept and return namedtuples?

functorch/_src/vmap.py
```diff
@@ -22,6 +23,23 @@
...
+    for name in dir(torch.return_types):
+        if name.startswith('__'):
+            continue
+        attr = getattr(torch.return_types, name)
+        if inspect.isclass(attr):
+            return_type_class = attr
+            # Note: We capture the current `return_type_class` with default argument `constructor`
+            # in the lambda otherwise we will point to the last value of `return_type_class` for all lambdas
+            torch.utils._pytree._register_pytree_node(return_type_class, lambda x: (
+                tuple(x), None), lambda x, c, constructor=return_type_class: constructor(x))
```
zou3519, line 38: tuple(x) should be list(x): flatten_function are supposed to return List[Any]: https://github.com/pytorch/pytorch/blob/de902b5d02ef0b264d4ce1406b8e16c7ba1e2680/torch/utils/_pytree.py#L54-L55

functorch/_src/vmap.py
```diff
@@ -22,6 +23,23 @@
...
+
+def register_namedtuple_return_types():
+    # Register torch.return_types as pytree node.
+    for name in dir(torch.return_types):
+        if name.startswith('__'):
+            continue
+        attr = getattr(torch.return_types, name)
+        if inspect.isclass(attr):
+            return_type_class = attr
+            # Note: We capture the current `return_type_class` with default argument `constructor`
```
zou3519, line 35: very interesting

functorch/_src/vmap.py
```diff
@@ -22,6 +23,23 @@
 in_dims_t = Union[int, Tuple]
 out_dims_t = Union[int, Tuple[int, ...]]
 
+
+def register_namedtuple_return_types():
```
zou3519, line 27: nit: torch.return_types are not exactly namedtuples :(. They have different semantics from Python namedtuple but they behave the same; the right terminology is Struct Sequence (https://docs.python.org/3/c-api/tuple.html#struct-sequence-objects).

Let's name this something like "register_torch_return_types()"?

zou3519 (commented):
I had some minor nits around naming. We should check that regular Python namedtuples still work (I think they will). Other than those, this LGTM, and we should probably upstream this into pytorch/pytorch after so that other users of torch.utils._pytree can benefit.

functorch/_src/pytree_hacks.py
kshitij12345: I don't think there are operators which return `namedtuples`. I mostly refer to `torch.return_types` as `namedtuples` probably because the [PR](https://github.com/pytorch/pytorch/pull/15429) that added them referred them as such.

Do let me know if you know any such operator.
Thanks! :)

zou3519: Hmm my suggestion was to test vmap over a function that returns a namedtuple. The function doesn't have to be an operator:
```
Point = namedtuple('Point', ['x', 'y'])

def f(x, y):
  return Point(x=x, y=y)

x = torch.randn(2, 5)
y = torch.randn(2, 3)
result = vmap(f)(x, y)
# check that result is a Point
```

kshitij12345: Updated.  Thanks :)!

zou3519 (approved):
LGTM! We should upstream the relevant pytree parts of this into torch.utils._pytree

PR #358, opened 2021-12-20
Add wheel builds

malfet:


.github/workflows/wheels.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: actions/checkout@v2
+      - name: Install PyTorch Nightly
+        run: |
+          python3 -mpip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+      - name: Build wheel
+        run: |
+          python3 -mpip install wheel
+          python3 setup.py bdist_wheel
+      - name: Upload wheel as GHA artifact
+        uses: actions/upload-artifact@v2
```
zou3519, line 31: @malfet where does this end up uploading to?

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,34 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request
+
+jobs:
+
+  build-wheel:
+    runs-on: ubuntu-18.04
```
zou3519, line 11: Do you have guidance for how to do builds on Mac or Windows?

.github/workflows/wheels.yml
malfet: To projects actions page, for example https://github.com/pytorch/functorch/actions/runs/1604493500

.github/workflows/wheels.yml
malfet: Let me extend the matrix to Mac, not sure about Windows though

malfet: Done

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,36 @@
+name: Wheels
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - main
```
zou3519, line 7: I think this file makes sense to me. When we are ready to cut a release binary against PyTorch 1.11, would we create a separate functorch branch (e.g. release/0.1.0) and modify this file to install the PyTorch RC?

zou3519 (approved):
This is pretty cool, thanks @malfet!

Looking at the CI signal you might need to rebase this PR on top of the main branch

.github/workflows/wheels.yml
malfet: Good point, let me add `- release/*` wildcard, and tags one as well

.github/workflows/wheels.yml
```diff
@@ -0,0 +1,36 @@
+name: Wheels
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
vfdev-5, line 4: @zou3519 why we need to execute this workflow (upload wheels) on each PR ? Maybe I'm missing something ...

zou3519: I think the upload wheels only uploads the wheels to github actions. We technically don't need to do this on each PR if we want to save bandwidth but it is still a good check that the wheel building command runs without problem

PR #361, opened 2021-12-21
Fix parameter declarator cannot be qualified error

malfet:
Before this change `include_guard` declaration can be interpreted as both local variable declaration and function prototype
Disambiguate it by using curly bracket


malfet:
Discovered while trying to enable Mac wheels build in https://github.com/pytorch/functorch/pull/358

zou3519 (approved):
thanks!

bdhirsh (approved):

PR #375, opened 2022-01-04
README.md: don't include prompt in example code snippets

t-vi:
The >>> is rather obnoxious when copy-pasting the examples.


zou3519 (approved):
Thanks!

PR #383, opened 2022-01-05
[Compile Cache] Handle non tensor args

anijain2305:
The non tensor args are hashed in python and sent to the cache. Currently, all the non tensor args are cached.

@Chillee

### DIFF
--- a/test/test_compile_cache.py
+++ b/test/test_compile_cache.py
@@ -171,64 +171,117 @@ class TestCompileCache(TestCase):
 
 
 class TestCompileCacheNonTensorArgs(TestCase):
-    def check(self, a, b, mem_optimized_fn, fn):
+    def check(self, a, b, aot_autograd_fn, fn):
         a_clone = a.clone().detach().requires_grad_(True)
         ref = fn(a, b)
         ref.sum().backward()
 
-        res = mem_optimized_fn(a_clone, b)
+        res = aot_autograd_fn(a_clone, b)
         res.sum().backward()
         assert torch.allclose(res, ref)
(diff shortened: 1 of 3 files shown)


test/test_compile_cache.py
```diff
@@ -215,6 +215,88 @@ def _nop_compile(x, _):
...
+        for hasher_type in ["DynamicShapeHasher", "StaticShapeHasher"]:
+            functorch.compile.clear_compile_cache()
+
+            def _nop_compile(x, _):
+                return x
+
+            start_num_recomps = functorch.compile.num_of_recompilations()
+
+            aot_autograd_f = compiled_function(
+                fn, _nop_compile, _nop_compile, hasher_type=hasher_type
```
Chillee, line 244: there's a `nop` compiler in `compilers.py` that you should be able to import

test/test_compile_cache.py
```diff
@@ -215,6 +215,88 @@ def _nop_compile(x, _):
...
+
+        def fn_dropout(x, prob):
+            return torch.nn.functional.dropout(x, p=prob)
+
+        def check(a, b, mem_optimized_fn, fn):
+            a_clone = a.clone().detach().requires_grad_(True)
+            ref = fn(a, b)
+            ref.sum().backward()
+
+            res = mem_optimized_fn(a_clone, b)
```
Chillee, line 230: should probably be called something else since it's not necessary a mem_optimized_fn?

Chillee (approved):
LGTM mostly!

Can you do some benchmarking as well?

Also, add some more tests, including with a custom object with a `__hash__` defined on it.

PR #386, opened 2022-01-10
Added support for linalg.lu_factor* as _lu_with_info will be deprecated

vfdev-5:
Description:
- following https://github.com/pytorch/functorch/issues/240#issuecomment-1007489523 adding batch rule for replacing ops and keeping `_lu_with_info`


vfdev-5:
Some failures may be related to https://github.com/pytorch/pytorch/pull/69437  ... (thanks @Lezcano for pointing out the PR)

functorch/csrc/BatchRulesLinearAlgebra.cpp
```diff
@@ -185,6 +185,8 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
 
   VARIADIC_BDIMS_BOXED(_det_lu_based_helper);
   VARIADIC_BDIMS_BOXED(_lu_with_info);
+  VARIADIC_BDIMS_BOXED(linalg_lu_factor_ex);
+  VARIADIC_BDIMS_BOXED(linalg_lu_factor);
```
vfdev-5, line 189: This is wrong as the op should be decomposed.

PR #388, opened 2022-01-10
Cleanup for memory efficient fusion

anijain2305:
Cleanup


anijain2305:
CI fails with a known error

PR #390, opened 2022-01-10
Minor refactor for default partition.

anijain2305:
As Title

### DIFF
--- a/functorch/_src/aot_autograd.py
+++ b/functorch/_src/aot_autograd.py
@@ -125,8 +125,8 @@ def _extract_fwd_bwd_modules(joint_module: fx.GraphModule, saved_values):
     fwd_outputs, bwd_outputs = _extract_fwd_bwd_outputs(joint_module)
     primal_inputs = list(filter(_is_primal, joint_module.graph.nodes))
     tangent_inputs = list(filter(_is_tangent, joint_module.graph.nodes))
-    # Construct the forward module
-    fwd_graph = _extract_graph_with_inputs_outputs(joint_module.graph, primal_inputs, fwd_outputs + saved_values)
+
+    # Construct the backward graph to filter out unused saved values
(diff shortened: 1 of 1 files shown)


functorch/_src/aot_autograd.py
```diff
@@ -151,10 +151,16 @@ def default_partition(joint_module: fx.GraphModule, _joint_inputs):
     primal_inputs = list(filter(_is_primal, joint_module.graph.nodes))
     fwd_outputs, bwd_outputs = _extract_fwd_bwd_outputs(joint_module)
     forward_only_graph = _extract_graph_with_inputs_outputs(joint_module.graph, primal_inputs, fwd_outputs)
-    forward_node_names = set([node.name for node in forward_only_graph.nodes if node.op != 'output'])
+    forward_node_names = set([node.name for node in forward_only_graph.nodes])
```
Chillee, line 154: What's the point of this change?

anijain2305: Goal of the PR is to find the boundary ops (instead of relying on the generation of bwd pass later to find out which inputs are unused). To do this, I go through each node in the joint graph, and collect the ones who are in fwd graph but atleast one of its users are in bwd pass. 

Therefore, I need to collect the `output` node as well in the forward. For a graph like following

~~~
def forward(self, primals_1, primals_2):
    mm = torch.ops.aten.mm(primals_1, primals_2)
    neg = torch.ops.aten.neg(mm);  mm = None
    neg_1 = torch.ops.aten.neg(neg);  neg = None
    neg_2 = torch.ops.aten.neg(neg_1);  neg_1 = None
    return [neg_2, primals_1, primals_2]
~~~

If `output` op is not saved in the `forward_node_names`, then it leads to three saved outputs `[primals_1, primals_2, neg_2]` when it should be only two `[primals_1, primals_2]`.

(This is not a problem today because later we generate bwd graph twice and remove the ones that are unused and `neg_2` is therefore removed there. It works fine, but maybe we should clean that up as well).

Chillee (approved):
Makes sense to me.

PR #397, opened 2022-01-12
Fixes printing issue with TensorWrapper

### LINKED ISSUE #382: Printing tensor broken in vmap(grad( case
Repro:
```
import torch
import torch.nn as nn
from functorch import vmap, grad, make_functional

class TestNetwork(nn.Module):
  def __init__(self):
    super().__init__()
    self.net = nn.Linear(5, 5)
  def forward(self, x):
    x = self.net(x)
    print(x)
    print(x.shape)
    print("Should be (10,5)")
    x[:, 0] += 3
    return x

model = TestNetwork()
data = torch.randn(10,5)
targets = torch.randn(10, 5)

func_model, params = make_functional(model)

def compute_loss(params, data, targets):
    preds = func_model(params, data.unsqueeze(0))
    return torch.mean((preds - targets.unsqueeze(0)) ** 2)

# This seems fine.
grad(compute_loss)(params, data, targets)

# This breaks
per_sample_grads = vmap(grad(compute_loss), (None, 0, 0))(params, data, targets)
```
gives `RuntimeError: Cannot access data pointer of Tensor that doesn't have storage`

vfdev-5:
Description:
- Hacky fix to fix printing tensors inside `grad`
- other nits and cosmetics

Fixes #382

@zou3519 what do you think ?

### DIFF
--- a/functorch/__init__.py
+++ b/functorch/__init__.py
@@ -91,14 +91,16 @@ def _functorch_str(tensor):
         return _old_str(tensor)
 
     value = _C.get_unwrapped(tensor)
-    # Disable temporarily kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
     dl_enabled = _C.tls_set_is_included()
-    if (dl_enabled):
-        _C._set_dynamic_layer_keys_included(False)
-    value_repr = repr(value)
-    # Reenable kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
-    if (dl_enabled):
-        _C._set_dynamic_layer_keys_included(True)
+    try:
+        # Disable temporarily kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+        if (dl_enabled):
+            _C._set_dynamic_layer_keys_included(False)
+        value_repr = repr(value)
+    finally:
+        # Reenable kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+        if (dl_enabled):
+            _C._set_dynamic_layer_keys_included(True)
 
     if _C.is_batchedtensor(tensor):
         bdim = _C.maybe_get_bdim(tensor)
--- a/functorch/__init__.py
+++ b/functorch/__init__.py
@@ -91,7 +91,15 @@ def _functorch_str(tensor):
         return _old_str(tensor)
 
     value = _C.get_unwrapped(tensor)
+    # Disable temporarily kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+    dl_enabled = _C.tls_set_is_included()
+    if (dl_enabled):
+        _C._set_dynamic_layer_keys_included(False)
     value_repr = repr(value)
+    # Reenable kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+    if (dl_enabled):
+        _C._set_dynamic_layer_keys_included(True)
+
     if _C.is_batchedtensor(tensor):
         bdim = _C.maybe_get_bdim(tensor)
         assert bdim != -1
--- a/test/test_eager_transforms.py
+++ b/test/test_eager_transforms.py
@@ -720,22 +720,22 @@ class TestGradTransform(TestCase):
         x = torch.tensor(3.14, device=device)
         functorch.grad(foo)(x)
 
-    @onlyCPU
     def test_tensor_print(self, device):
-        x = torch.tensor(3.14, device=device)
+        x = torch.tensor([[3.14, ], [3.14, ]], device=device)
         buf = None
 
-        def foo(x):
+        def foo(t):
             nonlocal buf
-            buf = repr(x)
-            return x
+            buf = repr(t)
+            return t.mean()
 
-        grad(grad(foo))(x)
-        expected = textwrap.dedent("""\
-                GradTrackingTensor(lvl=3, value=
-                    GradTrackingTensor(lvl=2, value=
-                        tensor(3.1400)
-                    )
+        device_info = "" if device == "cpu" else ", device='cuda:0'"
+
+        grad(foo)(x)
+        expected = textwrap.dedent(f"""\
+                GradTrackingTensor(lvl=2, value=
+                    tensor([[3.1400],
+                            [3.1400]]{device_info})
                 )""")
         self.assertEqual(buf, expected)
 
--- a/test/test_eager_transforms.py
+++ b/test/test_eager_transforms.py
@@ -10,7 +10,6 @@ from torch.testing._internal.common_utils import (
 import torch
 import torch.nn as nn
 import torch.nn.functional as F
-import textwrap
 import unittest
 import warnings
 import math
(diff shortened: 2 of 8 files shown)


functorch/__init__.py
```diff
@@ -91,7 +91,15 @@ def _functorch_str(tensor):
...
     value = _C.get_unwrapped(tensor)
+    # Disable temporarily kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+    dl_enabled = _C.tls_set_is_included()
+    if (dl_enabled):
+        _C._set_dynamic_layer_keys_included(False)
     value_repr = repr(value)
+    # Reenable kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+    if (dl_enabled):
+        _C._set_dynamic_layer_keys_included(True)
+
```
zou3519, line 102: Ugh I think I introduced this bug last month. I agree on the approach (that when we're printing the repr we should disable all things functorch related). Let me think about this a little more and get back to you

zou3519, line 102: @vfdev-5 I'm a bit confused about why our current tests pass ([1](https://github.com/pytorch/functorch/blob/b2d176daec0cfaee7d047a77d63cfc1ef020e057/test/test_vmap.py#L986-L1002), [2](https://github.com/pytorch/functorch/blob/main/test/test_eager_transforms.py#L723-L740)). Do you have some insight into that?

functorch/__init__.py
```diff
@@ -91,7 +91,15 @@ def _functorch_str(tensor):
...
 
     value = _C.get_unwrapped(tensor)
+    # Disable temporarily kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+    dl_enabled = _C.tls_set_is_included()
+    if (dl_enabled):
+        _C._set_dynamic_layer_keys_included(False)
     value_repr = repr(value)
+    # Reenable kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+    if (dl_enabled):
+        _C._set_dynamic_layer_keys_included(True)
```
zou3519, line 101: We probably want to use some "RAII" here: if the `repr` call fails then we want to make sure the state is still OK:
```
try: 
  if (dl_enabled):
        _C._set_dynamic_layer_keys_included(False)
finally:
   if (dl_enabled):
        _C._set_dynamic_layer_keys_included(True)
```

functorch/__init__.py
zou3519: I see, the problem is the printing path does different things on large tensor vs small tensors

zou3519 (commented):
Approach LGTM. One comment about RAII. We should also beef up our testing of printing and test some more cases :) the previous tests I wrote are not comprehensive at all.

functorch/__init__.py
zou3519: I thought about it and it looks like this was a pre-existing condition (and not caused by my refactor of DynamicLayer last month)

vfdev-5: Yes, I'm also surprised that test_eager_transforms.py::test_tensor_print was passing before this PR. I'll check that.
As for vmap, IMO they should always pass as it does not modify dispatch keys (tls_set_dispatch_key_included).

functorch/__init__.py
vfdev-5: I see your point, but just wonder if we really need to do that. If `value_repr = repr(value)` fails then script execution stops. If user does try-except some part of his/her code where `print(t)` errors, catches the error and continue the script, we should ensure that the state is OK and your pattern would totally make sense. I wonder how "typical" this pattern in python development ?

zou3519: > if user does try-except some part of his/her code where print(t) errors, catches the error and continue the script, we should ensure that the state is OK and your pattern would totally make sense.  I wonder how "typical" this pattern in python development ?

You're right that this pattern would probably not be common in user code. There's two concrete use cases for this that I've seen:
1. Someone has a long-running training loop. If an error occurs, I've seen some people do a try-catch and then save their model state in the finally() step so that no training progress is lost.
2. Testing. If someone (really, just us :P) has a test suite that tests tensor printing, if the printing raises an error in a unit test then it'll affect the state of all other subsequent unit tests

functorch/__init__.py
vfdev-5: Seems like there is no printing issue for scalars but there is an issue with non-scalar tensors.

test/test_eager_transforms.py
```diff
@@ -720,22 +720,22 @@ def foo(x):
         x = torch.tensor(3.14, device=device)
         functorch.grad(foo)(x)
 
-    @onlyCPU
     def test_tensor_print(self, device):
-        x = torch.tensor(3.14, device=device)
+        x = torch.tensor([[3.14, ], [3.14, ]], device=device)
```
zou3519, line 724: Could we keep the original tests and add some new test cases? It's probably valuable to test a variety of tensors to make sure they can be printed:
- torch.tensor(3.14, device=device)
- torch.tensor([[3.14, ], [3.14, ]], device=device)
- Some very big tensor (`torch.randn(64, 3, 32, 32)`). We can probably compare that `repr(t_inside_transform)` is `GradTrackingTensor(lvl=2, value={repr(t)}` or something

zou3519 (commented):
The code LGTM, but we should try to add some more rigorous tests. I suggested adding some cases that include the cross product of { different tensors } x { vmap, grad, vmapxvmap, gradxgrad, vmapxgrad } transforms

test/test_eager_transforms.py
zou3519: There are probably other edge cases that we want to test that I'm not thinking of here

test/test_eager_transforms.py
```diff
@@ -720,24 +719,43 @@ def foo(x):
...
-                )""")
-        self.assertEqual(buf, expected)
+    @parametrize("op_list_data", [
+        subtest(([vmap, ], [(64, 3, 32, 32), (4, 2)]), name='vmap'),
+        subtest(([vmap, vmap], [(64, 3, 32, 32), (4, 3, 2)]), name='vmap_vmap'),
+        subtest(([grad, ], [(0, ), [], (4, 2), (64, 3, 32, 32)]), name='grad'),
+        subtest(([grad, grad], [[], ]), name='grad_grad'),
+        subtest(([vmap, grad], [(4, 2)]), name='vmap_grad'),
+    ])
+    def test_tensor_print(self, device, op_list_data):
```
vfdev-5, line 729: @zou3519 how about that ? We can not make data x op product as data depends on the op.

test/test_vmap.py
```diff
@@ -984,24 +983,6 @@ def f(x, y):
...
-            return x
-
-        vmap(vmap(foo))(x)
-        expected = textwrap.dedent("""\
-                BatchedTensor(lvl=3, bdim=0, value=
-                    BatchedTensor(lvl=2, bdim=0, value=
-                        tensor([[3.1400]])
-                    )
-                )""")
-        self.assertEqual(buf, expected)
```
zou3519, line 1003: Let's add a test for different bdims here (to make sure we didn't hardcode bdim=0 lol)

test/test_eager_transforms.py
```diff
@@ -720,24 +719,43 @@ def foo(x):
...
+
+        op_list, shapes = op_list_data
+        data = [torch.randn(s, device=device) for s in shapes]
```
zou3519, line 732: Can we also test different dtypes somehow?

zou3519 (approved):
Some minor comments about testing different dtypes / bdims. This looks pretty good though

test/test_eager_transforms.py
zou3519: This looks great, thanks!

vfdev-5:
@zou3519 I added bdim and dtypes in the tests. Anything else do you think we can need of ?

zou3519 (approved):
Nah this looks great, thank you!

PR #398, opened 2022-01-13
Use num_bytes instead of numel in min-cut partitioning

anijain2305:
As Title

@Chillee


Chillee (approved):
LGTM!

I wonder whether we should error out or just choose a default...

anijain2305:
Let's error out for now. We can keep adding it as we test more models. And once a steady state is reached, we can default to `float`.

PR #399, opened 2022-01-14
Decompose aten std op

anijain2305:
As Title

### DIFF
--- a/functorch/_src/decompositions.py
+++ b/functorch/_src/decompositions.py
@@ -284,9 +284,7 @@ def var_decomposition(x, dims, correction=0, keepdim=False):
     if dims is None:
         dims = []
 
-    if isinstance(dims, int):
-        n = x.shape[dims]
-    elif isinstance(dims, (tuple, list)) and len(dims) == 0:
+    if isinstance(dims, (tuple, list)) and len(dims) == 0:
         n = x.numel()
     else:
         n = 1
@@ -306,6 +304,4 @@ def var_decomposition(x, dims, correction=0, keepdim=False):
 
 @register_decomposition(aten.std)
 def std_decomposition(x, dims, correction=0, keepdim=False):
-    if dims is None:
-        dims = []
     return aten.sqrt(aten.var(x, dims, correction=correction, keepdim=keepdim))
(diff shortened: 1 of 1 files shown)


anijain2305:
@Chillee This is ready for review.

functorch/_src/decompositions.py
```diff
@@ -277,3 +277,35 @@ def _s_where_canonicalization(a, b, c):
...
 def detach_decomposition(x):
     return x
+
+
+@register_decomposition(aten.var)
+def var_decomposition(x, dims, correction=0, keepdim=False):
+    if dims is None:
+        dims = []
+
+    if isinstance(dims, int):
```
Chillee, line 287: When does this happen?

functorch/_src/decompositions.py
```diff
@@ -277,3 +277,35 @@ def _s_where_canonicalization(a, b, c):
 @register_decomposition(aten.detach)
 def detach_decomposition(x):
     return x
+
+
+@register_decomposition(aten.var)
+def var_decomposition(x, dims, correction=0, keepdim=False):
+    if dims is None:
```
Chillee, line 284: Same for this.

Chillee (commented):
In general, we want to keep our decompositions as straightforward as possible.

functorch/_src/decompositions.py
anijain2305: So, this can happen because of this

https://github.com/pytorch/pytorch/blob/64670e414eb4bcba7dcc382bb16be1492cdd6e95/aten/src/ATen/native/native_functions.yaml#L4704-L4708

where dim arg is Optional and can be None.

But, when we later call mean and use dim, it can't handle NoneType 

https://github.com/pytorch/pytorch/blob/64670e414eb4bcba7dcc382bb16be1492cdd6e95/aten/src/ATen/native/native_functions.yaml#L2985-L2990

functorch/_src/decompositions.py
anijain2305: This can be removed.

Chillee:
I think we can probably simplify these more once we have overloads.

Ideally, these decompositions should be simple enough to be torchscriptable :P

But LGTM.

Chillee (approved):

Issue #381, opened 2022-01-05
Calling VMap on Grad - Issues with Indexing


bchen0:
Hi,

When calling vmap on grad in order to compute per-sample gradients, it seems that indexing breaks. See below for an example - after x = self.net(x), x is a "GradTrackingTensor" whose shape should be (10, 5) but instead has shape 5 (and as a result the indexing later breaks)

This only seems to happen when calling vmap on grad. Calling grad by itself doesn't seem to break.

```
class TestNetwork(nn.Module):
  def __init__(self):
    super().__init__()
    self.net = nn.Linear(5, 5)
  def forward(self, x):
    x = self.net(x)
    print(x.shape)
    print("Should be (10,5)")
    x[:, 0] += 3
    return x

model = TestNetwork()
data = torch.randn(10,5)
targets = torch.randn(10, 5)

func_model, params = make_functional(model)

def compute_loss(params, data, targets):
    preds = func_model(params, data)
    return torch.mean((preds - targets) ** 2)

# This seems fine.
grad(compute_loss)(params, data, targets)

# This breaks
per_sample_grads = vmap(grad(compute_loss), (None, 0, 0))(params, data, targets)
```

Thanks,
Brian

zou3519:
Thanks for the issue, @bchen0!

I'm assuming you have a batch size of 10 and want to compute a gradient per element in your batch (so, that is 10 gradients). If that is the case, then this is intentional - to compute per-sample-gradients with vmap, one should first write a function that computes gradients for a single example and then we can use vmap to lift it to computing a gradient per example.

More concretely, this means we need to modify compute_loss to handle `data` of shape (5,) and targets of shape (5,). The easiest way to do this without changing your model (which presumably assumes data and target have a batch dimension) is to add unsqueeze statements that add an additional dimension:

```
class TestNetwork(nn.Module):
  def __init__(self):
    super().__init__()
    self.net = nn.Linear(5, 5)
  def forward(self, x):
    x = self.net(x)
    print(x.shape)
    print("Should be (10,5)")
    x[:, 0] += 3
    return x

model = TestNetwork()
data = torch.randn(10,5)
targets = torch.randn(10, 5)

func_model, params = make_functional(model)

def compute_loss(params, data, targets):
    preds = func_model(params, data.unsqueeze(0))
    return torch.mean((preds - targets.unsqueeze(0)) ** 2)

# This seems fine.
grad(compute_loss)(params, data, targets)

# This breaks
[...]

PR #401, opened 2022-01-20
fix vmap : index_put_ (handle broadcasting of value)

kshitij12345:
Fixes: https://github.com/pytorch/functorch/issues/381
Handle broadcasting of `values` if the dimension shape after indexing is higher.

TODO : 
~~File an issue for code duplication~~

### DIFF
--- a/functorch/csrc/BatchRulesScatterOps.cpp
+++ b/functorch/csrc/BatchRulesScatterOps.cpp
@@ -9,6 +9,9 @@
 #include <ATen/Operators.h>
 #include <functorch/csrc/PlumbingHelper.h>
 #include <functorch/csrc/BatchedFallback.h>
+#include <ATen/native/TensorAdvancedIndexing.h>
+#include <ATen/native/IndexKernel.h>
+#include <ATen/native/IndexingUtils.h>
 
 
 namespace at { namespace functorch {
@@ -112,6 +115,57 @@ Tensor index_plumbing(const Tensor & self, const List<optional<Tensor>> & indice
 }
 
 namespace {
+  static VmapDimVector indexed_shape(const Tensor &src, TensorList indices_list)
+  {
+    int64_t dims_before = 0, dims_after = 0, dims_indexed = 0;
+    IntArrayRef replacement_shape;
+    for (const auto dim : c10::irange(indices_list.size())) {
+      if (!indices_list[dim].defined()) {
+        if (dims_indexed == 0) {
+          dims_before++;
+        } else {
+          dims_after++;
+        }
+      } else {
+        dims_indexed++;
+        replacement_shape = indices_list[dim].sizes();
+      }
+    }
+
+    // Replace indexed dimensions in src with stride 0 and the size of the result tensor.
+    // The offset in these dimensions is computed by the kernel using the index tensor's
+    // values and the stride of src. The new shape is not meaningful. It's used to make
+    // the shape compatible with the result tensor.
+    auto shape = VmapDimVector(src.sizes());
+    int64_t end = dims_before + dims_indexed;
+    shape.erase(shape.begin() + dims_before, shape.begin() + end);
+    shape.insert(shape.begin() + dims_before, replacement_shape.begin(), replacement_shape.end());
+    return shape;
+  }
+
+  static VmapDimVector get_indexed_shape(Tensor self, const torch::List<c10::optional<at::Tensor>> &orig)
+  {
+    // first expand BoolTensor (masks) or ByteTensor (masks) into 1 or more LongTensors
+    auto indices = at::native::expandTensors(self, orig);
+    // next broadcast all index tensors together
+    try {
+      indices = at::expand_outplace(indices);
+    } catch (std::exception &e) {
+      TORCH_CHECK_INDEX(false, "shape mismatch: indexing tensors could not be broadcast together"
+                               " with shapes ");
+    }
+    // add missing null Tensors so that it matches self.dim()
+    while (indices.size() < static_cast<size_t>(self.dim())) {
+      indices.emplace_back();
+    }
+    // if the non-null indices are not all adjacent, transpose self and indices
+    // together so that they're adjacent at the front
+    if (!at::native::hasContiguousSubspace(indices)) {
+      std::tie(self, indices) = at::native::transposeToFront(self, indices);
+    }
+    return indexed_shape(self, indices);
+  }
+
   std::tuple<Tensor, std::vector<optional<Tensor>>, Tensor>
   index_put_batch_rule_helper(const Tensor &self,
                               optional<int64_t> self_bdim,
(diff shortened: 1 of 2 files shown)


functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -122,10 +161,10 @@ void index_put__batch_rule(
   if (!self_bdim.has_value()) {
     vmapIncompatibleInplaceError("index_put");
   }
-  auto self_ = moveBatchDimToFront(self, self_bdim);
-  auto values_ = moveBatchDimToFront(values, values_bdim);
-  TORCH_INTERNAL_ASSERT(indices.size() == indices_bdims.size());
-  std::vector<optional<Tensor>> indices_ = batchIndices(indices, indices_bdims, self_.size(0), self_bdim, values_bdim);
+  Tensor self_, values_;
+  std::vector<optional<Tensor>> indices_;
+  std::tie(self_, indices_, values_) = index_put_batch_rule_helper(
```
zou3519, line 166: `batchIndices` is used in the implementation of `index_batch_rule`. Is that one OK?

zou3519: Nvm I see what is going on now

functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -111,6 +111,45 @@ Tensor index_plumbing(const Tensor & self, const List<optional<Tensor>> & indice
...
+    auto values_logical_rank = rankWithoutBatchDim(values, values_bdim);
+    Tensor self_ = moveBatchDimToFront(self, self_bdim);
+    Tensor values_ = moveBatchDimToFront(values, values_bdim);
+    TORCH_INTERNAL_ASSERT(indices.size() == indices_bdims.size());
+    std::vector<optional<Tensor>> indices_ = batchIndices(indices, indices_bdims, self_.size(0), self_bdim, values_bdim);
+
+    // handle case when prefix unit dimensions are stripped in setitem.
+    // values of shape `[1, 1, 2]` becomes `[2]`
+    // This leads to problem if values was actually batched.
+    // so we re-append the unit dimensions after batch_dim.
```
zou3519, line 133: I'm really confused about why this happens, and I think the problem is it's unclear what the actual semantics of index_put_ are. I read your comment over at https://github.com/pytorch/functorch/issues/381#issuecomment-1017247434 -- do you have a sense of why it's OK for PyTorch to always change values of shape `[1, 1, 2]` to `[2]`?

functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -111,6 +111,45 @@ Tensor index_plumbing(const Tensor & self, const List<optional<Tensor>> & indice
...
+    auto self_logical_rank = rankWithoutBatchDim(self, self_bdim);
+    auto values_logical_rank = rankWithoutBatchDim(values, values_bdim);
+    Tensor self_ = moveBatchDimToFront(self, self_bdim);
+    Tensor values_ = moveBatchDimToFront(values, values_bdim);
+    TORCH_INTERNAL_ASSERT(indices.size() == indices_bdims.size());
+    std::vector<optional<Tensor>> indices_ = batchIndices(indices, indices_bdims, self_.size(0), self_bdim, values_bdim);
+
+    // handle case when prefix unit dimensions are stripped in setitem.
+    // values of shape `[1, 1, 2]` becomes `[2]`
+    // This leads to problem if values was actually batched.
```
zou3519, line 132: Okay I think I understand what the problem is. It doesn't seem like a problem specific to setitem -- you can also run into this by directly calling index_put_. Please let me know if this is correct:

The problem is that `tensor.index_put_(idxs, value)'s batching rule didn't handle broadcasting the value. If we have ```
tensor =  torch.zeros(3, 4)
value = torch.ones(2)
idxs = (torch.tensor([0]), torch.tensor([1, 2]))
```
and we do `tensor.index_put_(idxs, value)`, tensor[idxs] is a tensor of shape (1, 2) and value is a tensor of shape (2,). value broadcasts up to the size of tensor[idxs].

However, if the values is batched, we have to be very careful: if the value is a BatchedTensor(Tensor[B, 2]), then we must insert additional 1's before invoking `index_put_` on regular tensors so that the value broadcasts up to the size of tensor[idxs].

The way we're determining the (logical) rank of tensor[idxs] is that it is equal to the (logical) rank of tensor, therefore the solution is to put 1's into the value until it has the same (logical) rank as that of tensor.

File: functorch/csrc/BatchRulesScatterOps.cpp [hunk 4]
```diff
@@ -111,6 +111,45 @@ Tensor index_plumbing(const Tensor & self, const List<optional<Tensor>> & indice
...
+    {
+      auto values_sizes = values_.sizes();
+      auto batch_offset = values_bdim.has_value() ? 1 : 0;
+      VmapDimVector new_values_shape(values_sizes.begin(), values_sizes.end());
+      size_t diff = self_.dim() - values_sizes.size();
+      // insert the unit dims again after batch_dim (if it exists).
+      for (const auto idx : c10::irange(0, diff))
+      {
+        (void)idx;
+        new_values_shape.insert(new_values_shape.begin() + batch_offset, 1);
```
zou3519, line 144: this is not really good for performance. Let's say we have an array [2, 3] and we're planning on inserting three 1's in between:
`[2, 1, 3] -> [2, 1, 1, 3] -> [2, 1, 1, 1, 3]`
The first insertion causes the array to move the element 3 to the right. The next insertion causes the array to move [1, 3] to the right. The third insertion causes the array to move [1, 1, 3] to the right. This is O(n^2)!

We can make it O(n) by constructing a VmapDimVector and then filling it in from left-to-right:
`[2] -> [2, 1, 1, 1] -> [2, 1, 1, 1, 3]`

functorch/csrc/BatchRulesScatterOps.cpp
zou3519: If this sounds correct:
- we should rewrite the comment string here to make it clearer that this is what is going on
- we should add some explicit tests for index_put_ under vmap

kshitij12345: Ah. Right. That is indeed the case. Sorry I was confused. Have updated the code and added a two test cases.

Thanks :)

kshitij12345:
@zou3519 would like to this offline with you once.

zou3519:
Sure

kshitij12345:
@zou3519 PTAL :)

functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -111,6 +114,104 @@ Tensor index_plumbing(const Tensor & self, const List<optional<Tensor>> & indice
...
+    auto self_logical_rank = rankWithoutBatchDim(self, self_bdim);
+    auto values_logical_rank = rankWithoutBatchDim(values, values_bdim);
+    Tensor self_ = moveBatchDimToFront(self, self_bdim);
+    Tensor values_ = moveBatchDimToFront(values, values_bdim);
+    const auto batch_size = self_.size(0);
+    values_ = ensure_has_bdim(values_, values_bdim.has_value(), batch_size);
+    TORCH_INTERNAL_ASSERT(indices.size() == indices_bdims.size());
+    std::vector<optional<Tensor>> indices_ = batchIndices(indices, indices_bdims, self_.size(0), self_bdim, values_bdim);
+
+    auto indexed_shape = get_indexed_shape(self_, List<optional<Tensor>>(indices_));
```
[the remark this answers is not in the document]
zou3519: Okay reading the code it looks like there's not an easy way to avoid this, so let's just keep it as-is unless you see something easy we can do

zou3519 (commented):
This looks pretty good to me. Just some minor comments

kshitij12345:
Nothing to change. Let's merge it :) Thanks!

PR #403, opened 2022-01-21
[requirements] missing networkx

stas00:
trying the examples it requires `networkx`.

If it's an optional requirement, could we then perhaps have some sort of `extras` group like:

```
pip install -e .[all]
```
which will then get all the requirements, including the optional ones?
```
extras = {}
extras["all"] = ["networkx", ...]
setup(
    ...
    extras_require=extras,
```

e.g. you can see how it's being used [here](https://github.com/huggingface/transformers/blob/master/setup.py#L367)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ To do devel install:
 pip install -e .
 ```
     
-To install with optional dependencies, e.g. for AOT:
+To install with optional dependencies, e.g. for AOTAutograd:
     
 ```
 pip install -e .[aot]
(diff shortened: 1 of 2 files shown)


vfdev-5:
@stas00 yes, networkx is an optional requirement, i'd say. I think we shouldn't add it into `requirements` as in the PR. Maybe, `extra_require` could be a good way to handle that.

stas00:
done. also updated README.md to show how to do dev install and then the extras. can later expand into more groups.

Chillee (approved):
LGTM!

PR #404, opened 2022-01-21
Enabled pytree of tensors output for jvp, jacfwd

### LINKED ISSUE #268: {jvp, jacfwd} should accept functions that have pytree inputs and pytree outputs
Demonstration from JAX:

- [x] jvp should accept functions that take pytrees as inputs
```
import jax
import jax.numpy as jnp

def f(x):
    a, b = x
    return 2 * a + 3 * b

x = jnp.array([1., 1])
y = jnp.array([1., 1])
t = jnp.array([1., 1])
jax.jvp(f, ((x, y),), ((t, t),))
# (DeviceArray([5., 5.], dtype=float32), DeviceArray([5., 5.], dtype=float32))
```

- [ ] jvp should accept functions that return pytrees
```
import jax
import jax.numpy as jnp

def f(x):
    return {'left': 2 * x, 'right': 3 * x}

x = jnp.array([1., 1])
t = jnp.array([1., 1])
jax.jvp(f, (x,), (t,))
# ({'left': DeviceArray([2., 2.], dtype=float32),
#   'right': DeviceArray([3., 3.], dtype=float32)},
#  {'left': DeviceArray([2., 2.], dtype=float32),
#   'right': DeviceArray([3., 3.], dtype=float32)})
```
It's probably easiest to tackle these separately; but jvp should also accept functions that both accept and return pytrees.

- [x] jacfwd should accept functions that take pytrees as inputs (pre-req is jvp supporting the same)
- [ ] jacfwd should accept functions that return pytrees (pre-req is jvp supporting the same)

vfdev-5:
Description:
- Enabled pytree of tensors output for jvp, jacfwd
- Added tests

Now we can do
```python
import torch
import functorch as ft

def sum_parabola(x):
    res = (2 * x**2 + 3 * x + 4).sum()
    return res, {"res": res}

x = torch.tensor([1., 1.5])
t = torch.tensor([2., 2.5])
out = ft.jvp(sum_parabola, (x,), (t,))
print("-- out ft vjp sum_parabola: ", out)

x = jnp.array([1., 1.5])
t = jnp.array([2., 2.5])
out = jax.jvp(sum_parabola, (x,), (t,))
print("-- out jax jvp sum_parabola: ", out)


x = torch.tensor([1., 1.5])
out = ft.jacfwd(sum_parabola)(x)
print("-- out ft jacfwd sum_parabola: ", out)

x = jnp.array([1., 1.5])
out = jax.jacfwd(sum_parabola)(x)
print("-- out jax jvp sum_parabola: ", out)

> 
-- out ft vjp sum_parabola:  ((tensor(22.), {'res': tensor(22.)}), (tensor(36.5000), {'res': tensor(36.5000)}))
-- out jax jvp sum_parabola:  ((DeviceArray(22., dtype=float32), {'res': DeviceArray(22., dtype=float32)}), (DeviceArray(36.5, dtype=float32), {'res': DeviceArray(36.5, dtype=float32)}))
-- out ft jacfwd sum_parabola:  (tensor([7., 9.]), {'res': tensor([7., 9.])})
-- out jax jvp sum_parabola:  (DeviceArray([7., 9.], dtype=float32), {'res': DeviceArray([7., 9.], dtype=float32)})
```

Fixes #268

### DIFF
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -4,7 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Callable, Union, Tuple
+from typing import Any, Callable, Union, Tuple, List
 import torch
 from functools import partial, wraps
 import contextlib
@@ -243,14 +243,13 @@ def vjp(func: Callable, *primals, has_aux=False):
             primals_out = func(*diff_primals)
 
             if has_aux:
-                # TODO: Add a test for that once vjp/jvp both have has_aux option
-                if not isinstance(primals_out, tuple):
-                    raise TypeError("Function output should be a tuple if has_aux is True")
+                if not (isinstance(primals_out, tuple) and len(primals_out) == 2):
+                    raise TypeError("Function output should be a tuple: (output, aux) if has_aux is True")
                 primals_out, aux = primals_out
                 aux = _undo_create_differentiable(aux, level)
 
             flat_primals_out, primals_out_spec = tree_flatten(primals_out)
-            assert_non_empty_output(flat_primals_out, 'vjp(f)')
+            assert_non_empty_tensor_output(flat_primals_out, 'vjp(f, *primals)')
             flat_diff_primals, primals_spec = tree_flatten(diff_primals)
             results = _undo_create_differentiable(primals_out, level)
 
@@ -400,7 +399,6 @@ def jacrev(func: Callable, argnums: Union[int, Tuple[int]] = 0, *, has_aux=False
 
         # See NOTE: [Computing jacobian with vmap and vjp for multiple outputs]
         flat_output, output_spec = tree_flatten(output)
-        assert_non_empty_output(flat_output, 'jacrev(f, ...)(*args)')
 
         # NB: vjp already checks that all outputs are tensors
         # Step 1: Construct grad_outputs by splitting the standard basis
@@ -590,7 +588,7 @@ def noop():
     yield
 
 
-def assert_flat_tuple_of_tensors(elts, api, argname):
+def assert_flat_tuple_of_tensors(elts: Tuple[torch.Tensor, ...], api: str, argname: str) -> None:
     if not isinstance(elts, tuple):
         raise RuntimeError(
             f'{api}: Expected {argname} to be a tuple of Tensors, got {type(elts)}')
@@ -605,12 +603,22 @@ def assert_flat_tuple_of_tensors(elts, api, argname):
             f'{api}: Expected {argname} to be a non-empty tuple of Tensors.')
 
 
-def assert_non_empty_output(output, api):
+def assert_non_empty_tensor_output(output: List[Any], api: str) -> None:
     if output == [None] or len(output) < 1:
-        raise RuntimeError(f'{api}: Expected non-empty output of f')
+        raise RuntimeError(
+            f'{api}: Expected f to be a function that has non-empty output (got output = {output})'
+        )
+    for o in output:
+        if not isinstance(o, torch.Tensor):
+            raise RuntimeError(
+                f'{api}: expected f(*primals) to return only tensors'
+                f', got unsupported type {type(o)}'
+            )
 
 
-def assert_output_is_tensor_or_tensors(output, api):
+def assert_output_is_tensor_or_tensors(
+    output: Union[torch.Tensor, Tuple[torch.Tensor, ...]], api: str
+) -> None:
     if isinstance(output, torch.Tensor):
         return
     if not isinstance(output, tuple):
@@ -645,7 +653,10 @@ jvp_str = 'jvp(f, primals, tangents)'
 
 def safe_unpack_dual(dual, strict):
     if not isinstance(dual, torch.Tensor):
-        raise RuntimeError(f"Expected tensors, got unsupported type {type(dual)}")
+        raise RuntimeError(
+            f'{jvp_str}: expected f(*args) to return only tensors'
+            f', got unsupported type {type(dual)}'
+        )
 
     primal, tangent = fwAD.unpack_dual(dual)
     if tangent is None:
@@ -734,7 +745,7 @@ def jvp(func, primals, tangents, *, strict=False):
             duals = tree_unflatten(flat_duals, primals_spec)
             result_duals = func(*duals)
             result_duals, spec = tree_flatten(result_duals)
-            assert_non_empty_output(result_duals, jvp_str)
+            assert_non_empty_tensor_output(result_duals, jvp_str)
 
             primals_out, tangents_out = \
                 zip(*[safe_unpack_dual(dual, strict) for dual in result_duals])
@@ -847,7 +858,6 @@ def jacfwd(func, argnums=0):
 
         results = vmap(push_jvp)(basis)
         jac_outs, spec = tree_flatten(results)
-        assert_non_empty_output(jac_outs, 'jacfwd(f, ...)(*args)')
 
         jac_outs_ins = tuple(
             tuple(
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -41,7 +41,7 @@ def _undo_create_differentiable(inps, level=None):
         if isinstance(x, tuple):
             return tree_map(unwrap_tensors, tuple(x))
 
-        return x
+        raise RuntimeError(f"Expected tensors, got unsupported type {type(x)}")
 
     return tree_map(unwrap_tensors, inps)
 
@@ -641,10 +641,7 @@ jvp_str = 'jvp(f, primals, tangents)'
 
 def safe_unpack_dual(dual, strict):
     if not isinstance(dual, torch.Tensor):
-        # return dual itself and an empty tensor
-        # JAX returns array((b'',), dtype=[('float0', 'V')])
-        # for tangents in a similar case
-        return dual, torch.tensor([])
+        raise RuntimeError(f"Expected tensors, got unsupported type {type(dual)}")
 
     primal, tangent = fwAD.unpack_dual(dual)
     if tangent is None:
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -41,7 +41,7 @@ def _undo_create_differentiable(inps, level=None):
         if isinstance(x, tuple):
             return tree_map(unwrap_tensors, tuple(x))
 
-        raise RuntimeError(f"Expected tensors, got unsupported type {type(x)}")
+        return x
 
     return tree_map(unwrap_tensors, inps)
 
@@ -396,16 +396,8 @@ def jacrev(func: Callable, argnums: Union[int, Tuple[int]] = 0, *, has_aux=False
 
         # See NOTE: [Computing jacobian with vmap and vjp for multiple outputs]
         flat_output, output_spec = tree_flatten(output)
-        if len(flat_output) == 0:
-            raise RuntimeError(
-                'jacrev(f, ...)(*args): expected f to return at least one Tensor, '
-                'got no Tensors.')
-        for out in flat_output:
-            if isinstance(out, torch.Tensor):
-                continue
-            raise RuntimeError(
-                'jacrev(f, ...)(*args): expected f to only return Tensors as '
-                f'outputs, got {type(out)}')
+        assert_non_empty_output(flat_output, 'jacrev(f, ...)(*args)')
+
         # NB: vjp already checks that all outputs are tensors
         # Step 1: Construct grad_outputs by splitting the standard basis
         flat_output_numels = tuple(out.numel() for out in flat_output)
@@ -609,6 +601,11 @@ def assert_flat_tuple_of_tensors(elts, api, argname):
             f'{api}: Expected {argname} to be a non-empty tuple of Tensors.')
 
 
+def assert_non_empty_output(output, api):
+    if output == [None] or len(output) < 1:
+        raise RuntimeError(f'{api}: Expected non-empty output of f')
+
+
 def assert_output_is_tensor_or_tensors(output, api):
     if isinstance(output, torch.Tensor):
         return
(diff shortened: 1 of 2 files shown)


test/test_eager_transforms.py
```diff
@@ -1636,15 +1636,41 @@ def f(x):
...
+        for i in range(2):
+            assert isinstance(out[i], list)
+            assert isinstance(out[i][0], tuple) and \
+                isinstance(out[i][0][1], dict)
+
+        # Checks primals and tangents for int(out)
+        assert isinstance(out[0][1], int) and isinstance(out[1][1], torch.Tensor)
+        # Checks primals and tangents for out > 0
+        assert isinstance(out[0][2], torch.Tensor) and out[0][2].dtype == torch.bool \
+            and isinstance(out[1][2], torch.Tensor)
```
zou3519, line 1673: I think we should restrict jvp to only allow returns that are Tensors

vfdev-5: I was based on your comment from the issue: https://github.com/pytorch/functorch/issues/268#issuecomment-1009354325 and replicating JAX behaviour if possible. 
JAX does not allow to return any type of output neither, e.g. if returning a float it will raise an error, but user can return int and bool. I was trying to replicate this behaviour...

zou3519: Hmm, I wonder if we can find examples of how this is used

zou3519: Let me know if you end up finding examples where people do return int/bool from a function and then use it with the jvp transform.

functorch doesn't have parity with JAX on how scalars are handled. If the user wants to differentiate a scalar, for the other transforms (e.g. grad), we require them to use a scalar Tensor.

It should be safer for now to be consistent with our other transforms and require that the function return only Tensors -- it's easy to relax this constraint if users come to us asking to relax it. What are your thoughts?

functorch/_src/eager_transforms.py
```diff
@@ -41,7 +41,7 @@ def unwrap_tensors(x):
         if isinstance(x, tuple):
             return tree_map(unwrap_tensors, tuple(x))
 
-        raise RuntimeError(f"Expected tensors, got unsupported type {type(x)}")
+        return x
```
zou3519, line 44: This might change how grad/vjp handle outputs as well. I'm not sure we actually have tests for this

test/test_eager_transforms.py
vfdev-5: I'm OK with returning only tensors requirement. If we would like to expand that we may have to properly design and test it (and not in this PR).

As for int/bool support, maybe this jax PR is relevant: https://github.com/google/jax/pull/4039

functorch/_src/eager_transforms.py
vfdev-5: IMO, yes, I think as well that there is no test for this line

vfdev-5:
@zou3519 let's hold on a bit, there is something broken with jvp and probably not tested

functorch/_src/eager_transforms.py
```diff
@@ -243,12 +243,16 @@ def vjp(func: Callable, *primals, has_aux=False):
             primals_out = func(*diff_primals)
 
             if has_aux:
+                # TODO: Add a test for that once vjp/jvp both have has_aux option
+                if not isinstance(primals_out, tuple):
+                    raise TypeError("Function output should be a tuple if has_aux is True")
```
zou3519, line 248: Should we assert it is a tuple of size 2?

functorch/_src/eager_transforms.py
```diff
@@ -243,12 +243,16 @@ def vjp(func: Callable, *primals, has_aux=False):
             primals_out = func(*diff_primals)
 
             if has_aux:
+                # TODO: Add a test for that once vjp/jvp both have has_aux option
```
zou3519, line 246: Why does the test have to wait until later?

functorch/_src/eager_transforms.py
vfdev-5: yes, we can do that. I'll update the code

functorch/_src/eager_transforms.py
vfdev-5: I think I thought about jacrev/jacfwd as they are tested with a decorator. Yes, we can add vjp test here as well

functorch/_src/eager_transforms.py
```diff
@@ -609,6 +605,11 @@ def assert_flat_tuple_of_tensors(elts, api, argname):
             f'{api}: Expected {argname} to be a non-empty tuple of Tensors.')
 
 
+def assert_non_empty_output(output, api):
+    if output == [None] or len(output) < 1:
+        raise RuntimeError(f'{api}: Expected non-empty output of f')
```
zou3519, line 610: nit: We can make this more descriptive: `"Expected f to be a function that has non-empty output (got output = {output})`

functorch/_src/eager_transforms.py
```diff
@@ -243,12 +243,16 @@ def vjp(func: Callable, *primals, has_aux=False):
...
+                # TODO: Add a test for that once vjp/jvp both have has_aux option
+                if not isinstance(primals_out, tuple):
+                    raise TypeError("Function output should be a tuple if has_aux is True")
                 primals_out, aux = primals_out
                 aux = _undo_create_differentiable(aux, level)
 
-            results = _undo_create_differentiable(primals_out, level)
-            flat_diff_primals, primals_spec = tree_flatten(diff_primals)
             flat_primals_out, primals_out_spec = tree_flatten(primals_out)
+            assert_non_empty_output(flat_primals_out, 'vjp(f)')
```
zou3519, line 253: nit: `vjp(f, *primals)` is the syntax

functorch/_src/eager_transforms.py
```diff
@@ -609,6 +605,11 @@ def assert_flat_tuple_of_tensors(elts, api, argname):
             f'{api}: Expected {argname} to be a non-empty tuple of Tensors.')
 
 
+def assert_non_empty_output(output, api):
```
zou3519, line 608: nit: we should write in a comment that output must be a List (or use type annotations, but we haven't really been doing that).

functorch/_src/eager_transforms.py
```diff
@@ -643,6 +644,9 @@ def assert_non_empty_list_of_tensors(output, api, argname):
 
 
 def safe_unpack_dual(dual, strict):
+    if not isinstance(dual, torch.Tensor):
+        raise RuntimeError(f"Expected tensors, got unsupported type {type(dual)}")
```
zou3519, line 648: nit: Is this only called in jvp? Maybe we should be a bit more descriptive about where these tensors came from (e.g. "jvp(f, args, tangents): expected f(*args) to return only tensors, got ..."

functorch/_src/eager_transforms.py
vfdev-5: I can start with type annotations :)

functorch/_src/eager_transforms.py
```diff
@@ -842,9 +846,9 @@ def push_jvp(basis):
             return jvp_out
 
         results = vmap(push_jvp)(basis)
-        assert_output_is_tensor_or_tensors(results, 'jacfwd(f, ...)(*args)')
-
         jac_outs, spec = tree_flatten(results)
+        assert_non_empty_output(jac_outs, 'jacfwd(f, ...)(*args)')
```
zou3519, line 850: Is this already covered by jvp?

vfdev-5: Yes. I'll remove it as well. I also spotted that in another PR.

test/test_eager_transforms.py
```diff
@@ -1198,20 +1198,58 @@ def f(x):
...
+            expected_aux = {'a': -torch.diag(x.sin()), 'b': [torch.diag(1.0 + x.tan() ** 2)]}
+            self.assertEqual(aux, expected_aux)
 
-    @FIXME_jacrev_only
+    @jacrev_and_jacfwd
+    def test_outputs_can_any_pytree(self, device, jacapi):
+        x = torch.randn(2, 3, device=device)
+
+        for output in [None, ()]:
+            with self.assertRaisesRegex(RuntimeError, "Expected non-empty output"):
```
zou3519, line 1227: Does the error message say "jacfwd/jacrev" or "vjp/jvp"? (I don't think it matters too much which one it is -- ideally it says jacfwd/jacrev but vjp/jvp is fine for now and we can always. make the error message better)

test/test_eager_transforms.py
```diff
@@ -1198,20 +1198,58 @@ def f(x):
...
+                jacapi(lambda _: output)(x)
+
+        for output in [1, True, 12.2, "abc"]:
+            with self.assertRaisesRegex(RuntimeError, "Expected tensors, got unsupported type"):
```
zou3519, line 1231: Ditto here -- does the error message name an API?

test/test_eager_transforms.py
```diff
@@ -1653,25 +1691,39 @@ def test_inputs_are_tuples_of_tensors(self, device):
...
-            jvp(g, (x,), (t,))
-        with self.assertRaisesRegex(RuntimeError, "a Tensor or Tensors"):
-            jvp(h, (x,), (t,))
+        for output in [None, ()]:
+            with self.assertRaisesRegex(RuntimeError, "Expected non-empty output"):
+                jvp(lambda _: output, (x,), (t,))
+
+        for output in [1, True, 12.2, "abc"]:
+            with self.assertRaisesRegex(RuntimeError, "Expected tensors, got unsupported type"):
+                jvp(lambda _: output, (x,), (t,))
```
zou3519, line 1704: Does the error message mention jvp?

zou3519 (commented):
This looks pretty good, I only had some minor comments.

The testing was very comprehensive! (though it should probably also include tests for vjp).

vfdev-5:
> though it should probably also include tests for vjp

OK, let me add that as well

functorch/_src/eager_transforms.py
vfdev-5: Thinking more about that IMO it would be better either introduce additional arg e.g. `api: str` to specify from where it comes or catch that elsewhere. There is a similar issue with `_undo_create_differentiable` which is used by multiple functions and error message is not always clear...

test/test_eager_transforms.py
```diff
@@ -1198,20 +1247,58 @@ def f(x):
...
+    @jacrev_and_jacfwd
+    def test_outputs_can_any_pytree(self, device, jacapi):
+        x = torch.randn(2, 3, device=device)
+
+        for output in [None, ()]:
+            with self.assertRaisesRegex(RuntimeError, r"Expected f to be a function that has non-empty output"):
+                jacapi(lambda _: output)(x)
+
+        for output in [1, True, 12.2, "abc"]:
+            with self.assertRaisesRegex(RuntimeError, r"expected f\(\*primals\) to return only tensors"):
```
zou3519, line 1280: Because it's a regex, we can actually check that it mentions some API, like (I don't remember if this is the syntax) "(vjp|jvp).*: Expected f to ..."

functorch/_src/eager_transforms.py
zou3519: Maybe file an issue for the future?

functorch/_src/eager_transforms.py
```diff
@@ -594,7 +588,7 @@ def noop():
     yield
 
 
-def assert_flat_tuple_of_tensors(elts, api, argname):
+def assert_flat_tuple_of_tensors(elts: Tuple[torch.Tensor, ...], api: str, argname: str) -> None:
```
zou3519, line 591: elts should be typed as Any, right? It's not guaranteed that we'll be pass a Tuple of Tensors to this function. mypy (if it ran on this file) should end up complaining that we are passing things that may be statically known to not be tuples to this function.

functorch/_src/eager_transforms.py
```diff
@@ -609,7 +603,22 @@ def assert_flat_tuple_of_tensors(elts, api, argname):
...
+    for o in output:
+        if not isinstance(o, torch.Tensor):
+            raise RuntimeError(
+                f'{api}: expected f(*primals) to return only tensors'
+                f', got unsupported type {type(o)}'
+            )
+
+
+def assert_output_is_tensor_or_tensors(
+    output: Union[torch.Tensor, Tuple[torch.Tensor, ...]], api: str
```
zou3519, line 620: Should output be typed as Any?

functorch/_src/eager_transforms.py
```diff
@@ -243,12 +243,15 @@ def vjp(func: Callable, *primals, has_aux=False):
             primals_out = func(*diff_primals)
 
             if has_aux:
+                if not (isinstance(primals_out, tuple) and len(primals_out) == 2):
+                    raise TypeError("Function output should be a tuple: (output, aux) if has_aux is True")
```
zou3519, line 247: Should this be a TypeError? (I don't know). If so, we probably want to make sure we are consistent with this...

In a lot of other places where we complain about the input function being malformed we return RuntimeError right now I think

zou3519, line 247: Also: nit: Should we say "vjp(f, *primals): output of function `f` should be a ..." to make it clearer where the error is coming from? If you agree then we should also update the test to check for this.

zou3519 (commented):
Last couple of comments and then I think we're good to go!

functorch/_src/eager_transforms.py
vfdev-5: Actually, I do not see in the codebase where this method is used. 
IMO the purpose of typing hints is to show expected type. If somewhere in the codebase we used it on an unacceptable data (which would raise an error) and mypy would complain about that, it is good for us to see a bad usage of the method statically and not on the runtime. I assume that raising `RuntimeError` is not desired behaviour.

functorch/_src/eager_transforms.py
vfdev-5: Yes, that's true that in other places we raise RuntimeError for type mismatch. I can make here also RuntimeError to make the code consistent. Later we can switch to TypeError if needed such error granularity.

zou3519: Sounds good

zou3519 (approved):
LGTM

zou3519:
Test failures are unrelated

PR #409, opened 2022-01-24
Add support for randn and rand

samdow:
* #420
* #412
* #411
* #410
* __->__ #409

Updates `vmap` to accept a randomness parameter ('error', 'different', or 'same') that determines what the randomness behavior should be. Also adds `torch.randn` and `torch.rand` to demonstrate basic examples


test/test_vmap.py
```diff
@@ -3266,13 +3198,11 @@ def test_vmap_exhaustive(self, device, dtype, op):
         xfail('nn.functional.huber_loss'),
         xfail('nn.functional.instance_norm'),
         xfail('nn.functional.poisson_nll_loss'),
-        xfail('nn.functional.max_pool3d'),
```
samdow, line 3269: revert? unclear how this happened

functorch/_src/vmap.py
```diff
@@ -203,7 +203,7 @@ def _get_name(func: Callable):
 # on BatchedTensors perform the batched operations that the user is asking for.
 
 
-def vmap(func: Callable, in_dims: in_dims_t = 0, out_dims: out_dims_t = 0) -> Callable:
+def vmap(func: Callable, in_dims: in_dims_t = 0, out_dims: out_dims_t = 0, use_batched_random=True) -> Callable:
```
samdow, line 206: We should bikeshed this name...
I called it `use_batched_randomness` to mean that if True, each batch from vmap will have different random numbers. But I can definitely see a user thinking this meant that the randomness will be the same across batches (as I myself got confused while writing this sometimes)

Some potential jumping off points:
  - `use_batched_random`
  - `use_different_random`
  - `matching_random_across_batches`

functorch/csrc/init.cpp
```diff
@@ -252,17 +252,17 @@ namespace at { namespace functorch {
...
+  m.def("_grad_decrement_nesting", &at::functorch::_grad_decrement_nesting, "remove grad layer");
+  m.def("_wrap_for_grad", &at::functorch::_wrap_for_grad, "wrap tensor, used by grad transform");
+  m.def("_unwrap_for_grad", &at::functorch::_unwrap_for_grad, "unwrap tensor, used by grad transform");
   m.def("_set_vmap_fallback_warning_enabled", &at::functorch::setVmapFallbackWarningEnabled, "Set vmap fallback warnings");
   m.def("_set_vmap_fallback_enabled", &at::functorch::setVmapFallbackEnabled);
   m.def("_is_vmap_fallback_enabled", &at::functorch::isVmapFallbackEnabled);
-  m.def("dlevel", &at::functorch::dlevel, "dlevel");
-  m.def("dump_tensor", &at::functorch::dump_tensor, "dump_tensor");
+  m.def("dlevel", &at::functorch::dlevel, "");
+  m.def("dump_tensor", &at::functorch::dump_tensor, "dump tensor");
```
samdow, line 265: moving this to a separate change

functorch/_src/vmap.py
```diff
@@ -226,6 +230,11 @@ def vmap(func: Callable, in_dims: in_dims_t = 0, out_dims: out_dims_t = 0) -> Ca
         out_dims (int or Tuple[int]): Specifies where the mapped dimension
             should appear in the outputs. If :attr:`out_dims` is a Tuple, then
             it should have one element per output. Default: 0.
+        use_batched_random (str): Specifies whether the randomness in this
+            vmap should be the same or different across batches. If 'different',
+            the randomness for each batch will be different. If 'same', the
+            randomness will be the same across batches. If 'error', any calls to
+            random functions will error. Default: True.
```
zou3519, line 237: flag is called "randomness", not "use_batched_random"?

functorch/csrc/DynamicLayer.h
```diff
@@ -39,12 +41,14 @@ struct TORCH_API DynamicLayer {
   // Honestly these should be a union or some extendable metadata class.
   // Not doing that for now because I don't think we'll use this mechanism for very long.
   optional<int64_t> batchSize_;
+  optional<std::string> randomness_;
```
zou3519, line 44: This is a string in Python land but the usual way we represent string-like things in C++ are enums. enums save memory and dynamic allocations and are faster.

functorch/_src/vmap.py
zou3519: Also, I'd add a comment somewhere that this differs from JAX's randomness behavior

functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -0,0 +1,73 @@
...
+    "'same' or 'different' randomness flags on vmap or perform the randomness operation out of vmap"
+  );
+}
+
+template <typename F, F Func, typename... ExtraArgs>
+Tensor random_batching_rule(IntArrayRef shape, ExtraArgs... extra_args) {
+  c10::impl::ExcludeDispatchKeyGuard guard(kVmapModeKey);
+  auto maybe_layer = maybeCurrentDynamicLayer();
+  VmapDimVector shapeVec(shape.begin(), shape.end());
+  shapeVec.insert(shapeVec.begin(), maybe_layer->batchSize());
```
zou3519, line 27: insertion at the beginning is O(n), should start with [batch_size] and then insert shape.

zou3519 (commented):
Some comments for things in this PR that should probably be resolved:
1. std::string should be passed by reference, not value, otherwise it gets copied (and the copy is a dynamic allocation
2. We should really use an enum instead of a string in C++, that's how things like "Reduction::Mean" in PyTorch's Loss functions are implemented.

Some meta-comments that are not actionable:
1. Would it be better if we completed the feature and then cherry-picked all the commits to the release branch? There is a risk that we won't complete it by the release, right? So this would depend on how this feature behaves in a half-complete state. If it just errors out on everything that is not supported then we don't have to worry about cherry-picking everything together.
2. One thing that I have not thought very much about is: what if we decide later that we want a stateless rng API? Does this randomess API play well with that? (I'm not sure). Ideally we would leave both routes open, but we did decide to try out this version of the randomness API.

functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -0,0 +1,73 @@
...
+  std::string randomness = maybe_layer->randomness();
+  check_randomness(randomness);
+  if (randomness == "different") {
+    return makeBatched(Func(shapeVec, std::forward<ExtraArgs>(extra_args)...), 0, maybe_layer->layerId());
+  } else {
+    const auto res = Func(shape, std::forward<ExtraArgs>(extra_args)...);
+    return makeBatched(res.unsqueeze(0).expand(shapeVec), 0, maybe_layer->layerId());
```
zou3519, line 34: @samdow to improve performance this should return a regular Tensor, not a BatchedTensor with an expanded underlying Tensor.

Let's say we did an operation on the resulting Tensor, like x.sin(). If the tensor was expanded, we'll end up creating a new contiguous tensor and materializing all of the memory. If the Tensor wasn't expanded, then we save on some performance

zou3519: Ah I see, it was fixed in https://github.com/pytorch/functorch/pull/410/files. You read my mind, three weeks ago :)

samdow: Yeah this was actually a rebase error--I meant to change it in this one too 🤦‍♀️

PR #410, opened 2022-01-24
randint support

samdow:
* (to be filled)

### DIFF
--- a/functorch/csrc/BatchRulesRandomness.cpp
+++ b/functorch/csrc/BatchRulesRandomness.cpp
@@ -25,6 +25,34 @@ Tensor random_batching_rule(IntArrayRef shape, ExtraArgs... extra_args) {
     }
 }
 
+template <typename F, F Func, typename... ExtraArgs>
+Tensor randint_batching_rule(int64_t high, IntArrayRef shape, ExtraArgs... extra_args) {
+    c10::impl::ExcludeDispatchKeyGuard guard(kVmapModeKey);
+    auto maybe_layer = maybeCurrentDynamicLayer();
+    VmapDimVector shapeVec(shape.begin(), shape.end());
+    shapeVec.insert(shapeVec.begin(), maybe_layer->batchSize());
+    if (maybe_layer->useBatchedRandom()) {
+      return makeBatched(Func(high, shapeVec, std::forward<ExtraArgs>(extra_args)...), 0, maybe_layer->layerId());
+    } else {
+      const auto res = Func(high, shape, std::forward<ExtraArgs>(extra_args)...);
+      return makeBatched(res.unsqueeze(0).expand(shapeVec), 0, maybe_layer->layerId());
+    }
+}
+
+template <typename F, F Func, typename... ExtraArgs>
+Tensor randint_batching_rule(int64_t high, int64_t low, IntArrayRef shape, ExtraArgs... extra_args) {
+    c10::impl::ExcludeDispatchKeyGuard guard(kVmapModeKey);
+    auto maybe_layer = maybeCurrentDynamicLayer();
+    VmapDimVector shapeVec(shape.begin(), shape.end());
+    shapeVec.insert(shapeVec.begin(), maybe_layer->batchSize());
+    if (maybe_layer->useBatchedRandom()) {
+      return makeBatched(Func(high, low, shapeVec, std::forward<ExtraArgs>(extra_args)...), 0, maybe_layer->layerId());
+    } else {
+      const auto res = Func(high, low, shape, std::forward<ExtraArgs>(extra_args)...);
+      return makeBatched(res.unsqueeze(0).expand(shapeVec), 0, maybe_layer->layerId());
+    }
+}
+
 template <typename A, A a, typename C>
 struct RandomBatchRuleHelper;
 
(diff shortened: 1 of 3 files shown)


functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -67,7 +114,15 @@ TORCH_LIBRARY_IMPL(aten, FuncTorchVmapMode, m) {
   RANDOM_BATCH_RULE2(rand, generator_with_names);
   RANDOM_BATCH_RULE2(rand, names);
 
+  RANDINT_BATCH_RULE(randint);
+  RANDINT_BATCH_RULE2(randint, generator);
+  RANDINT_LOW_BATCH_RULE(randint, low);
+  RANDINT_LOW_BATCH_RULE(randint, low_generator);
```
zou3519, line 120: Hmm, there has to be a better way of doing this than doing macros + template hackery...

samdow: Sorry which part are you referring to? Like the fact that we have to write new macros for rand vs randint?

zou3519: I'm referring to the
```
template <typename A, A a, typename C>
struct RandIntBatchRuleHelper;

template <typename F, F Func, typename T1, typename T2, typename... T>
struct RandIntBatchRuleHelper<F, Func, typelist<T1, T2, T...>> {
  static Tensor apply(int64_t high, IntArrayRef shape, T... extra_args) {
    return random_batching_rule<decltype(&rand_int_wrapper<F, Func, T...>),
                                &rand_int_wrapper<F, Func, T...>,
                                int64_t, T...>(shape, high, std::forward<T>(extra_args)...);
  }
};
```
things going on

samdow: Still unclear--the existence of `RandIntBatchRuleHelper` or the use `rand_int_wrapper`?

zou3519: There seems to be a lot of boilerplate necessary to get the batching rule:
- we've got a RANDINT_BATCH_RULE macro
- which then calls a heavily templated RandIntBatchRuleHelper struct
- which then calls the templated random_batching_rule
- which has rand_int_wrapper as a template parameter.

Thinking about this a bit more relative to VMAP_SUPPORT, which is the "standard" way of defining a batching rule:
- there's some autogenerated plumbing that connects the vmap internals to a batching rule
- there's a "batching rule" which just defines the logic of what to do. The "batching rule" only involves regular Tensors and does not involve needing to peek at vmap internals.
The benefit of having the plumbing / batching rule abstraction is that (hopefully) we can make changes to the plumbing in one place and independently from the logic of the batching rule.

To apply that to what we have here, the "batching rule" would probably just be a function
```
std::tuple<Tensor,optional<int64_t>> randint(batch_size, randomness, usual_args...) {
  ...
}
```

And I'd expect the plumbing to do something like:
- get the current level
- Set up the ExcludeDispatchKeyGuard
- call the batching rule that is passed to the plumbing

That being said, I'm not sure what that would look like or if it would have as much boilerplate as is currently here

samdow: I see--thanks for the explanation

I don't think this is what you're saying but just for clarity (or in case others are confused), we can't use VMAP_SUPPORT here because that excludes the batching key while we want to run on the vmap mode key and need to exclude that. 
There might be some way to generalize so that it could take in a key to exclude and return randomness but that seemed like it had a high probability of breaking and I got intimidated by that. I also considered repeating something like the OutOfPlacePlumbing codegen for random but figured it would be confusing to have the codegen for every single function when it only applies to a couple dozen. I thought about trying to match on a set of names but trying to codegen for that seemed a little daunting. These definitely don't mean that we shouldn't do write codegen for this, just some background

I'm also not sure that codegen will eliminate the templates. Specifically, most of the templates are here because each function has a couple of variants (most have 2 that we deal with, with or without a generator, but randn has 4) that have different signatures after size. If we didn't have this generality of what could come after size, we wouldn't be able to do the albeit gross hack for randint since random_batching_rule is set up to take in sizes followed by any number of arguments (that will be passed directly through since sizes is the only thing that may need to be updated). So, without the generality on random_batching_rule, we would need N different functions for N different batching rules that are all the same batching rule (except for the inputs). In the start, I separated rand/randn from randint and randperm and honestly that amount of repetition made keeping even those three functions the same rather difficult (that's actually why this PR makes a change to random_batching_rule--I meant to do that in the earlier PR but it got messed up with stack and having repetition). This is why I ended up collapsing those three batching rules into one with the templating in RandIntBatchRuleHelper::apply, but we can definitely discuss or think that having multiple rules is okay

Ideally I would love to not have to repeat the rules as well as have some generic templating that's like "this function will have maybe some arguments before shape, then shape, then some arguments after shape" and write one batching rule for it. However, currently templating only supports one variadic arg and doesn't have that advanced amount of pattern matching. Similarly any thoughts, solutions, or input would be great

As a meta point, I also get that it's not looking great since the first 5 of these have to do all those steps that you outlined. The hope with these was to outline the "function shapes" we'll have to be looking at by doing (1) functions that only take scalars, (2) functions that take at least one tensor, (3) in-place functions. There's ~10 functions left to do and I think that we'll have to add 2 more templates, others are able to reuse ones from earlier. Still not great to have like 7 templates for 15 functions, but wanted to clarify that if a future user adds a new function that needs special randomness treatment, they (hopefully) won't have to go through all of this

PR #412, opened 2022-01-24
random_ and normal_ support

samdow:
* #523
* #521
* #420
* #511
* #510
* __->__ #412


test/test_vmap.py
```diff
@@ -3490,6 +3487,45 @@ def test_randperm(self, device, randomness, use_generator):
...
+        ]
+
+        B0 = 4
+        seed = 1234567
+        generator = torch.Generator(device=device)
+        orig_state = generator.get_state()
+        kwargs = {'generator': generator} if use_generator else {}
+
+        for op in supported_ops:
+            vmaped_value, unvmaped_value = torch.randn(B0, B0, device=device), torch.randn(B0, B0, device=device)
```
zou3519, line 3509: Why is it `torch.randn(B0, B0)`? "B0" is generally used in this file as the name for a dimension that gets vmapped over (i.e. a batch dimension). It would make the test a bit clearer if we did something like:

vmaped_value, unvmaped_value = torch.randn(B0, device=device), torch.randn(3, 4, device=device)

test/test_vmap.py
```diff
@@ -3490,6 +3487,45 @@ def test_randperm(self, device, randomness, use_generator):
...
+                    vmap(op, in_dims=(None, 0), randomness=randomness)(unvmaped_value, vmaped_value)
+                return
+
+            vmap(op, in_dims=(None, 0), randomness=randomness)(unvmaped_value, vmaped_value)
+
+            passed = torch.randn([B0, B0], device=device)  # reset for in place operation
+            generator = generator.set_state(orig_state)
+            torch.manual_seed(seed)
+            op(passed, vmaped_value)
+            assert torch.allclose(unvmaped_value, passed)
```
zou3519, line 3527: nit: prefer self.assertEqual instead of `assert torch.allclose`. self.assertEqual is hooked into the precision override system for PyTorch testing

test/test_vmap.py
```diff
@@ -3430,11 +3422,13 @@ def reset_random():
         generator = torch.Generator(device=device)
         orig_state = generator.get_state()
         kwargs = {'device': device, 'generator': generator} if use_generator else {'device': device}
+        only_gen_kwarg = {'generator': generator} if use_generator else {}
         supported_random_ops = [
             lambda _, shape: torch.randn(shape, **kwargs),
             lambda _, shape: torch.rand(shape, **kwargs),
             lambda _, shape: torch.randint(100, shape, **kwargs),
             lambda _, shape: torch.randint(5, 100, shape, **kwargs),
+            lambda t, _: t.random_(**only_gen_kwarg),
```
zou3519, line 3431: (observation, no action required) ack, it's a little weird that this function takes in two args and the others take in a single arg

zou3519 (commented):
The code looks good to me, I had some comments around improving the testing.

In general it might be better to separate out the tests? I would have expected:
(a) a test for all random functions that do not take tensor arguments
(b) a test for all random in-place functions that take a single Tensor argument

But it looks like we've taken (b) and split it up and are using (a) for half the logic

PR #413, opened 2022-01-25
[Compile Cache] Caching the forward and backward compiler ids

anijain2305:
For same function, we might want to benchmark different compilers. Previously, compile cache was not caching the compilers. This PR caches the ids of fwd and bwd compilers.

@Chillee


Chillee:
Looks reasonable to me.

PR #416, opened 2022-01-26
Fixed some test cases/added some easy batching rules

Chillee:


Chillee:
@lucidrains at least this one actually passes CI lmao

PR #419, opened 2022-01-27
Added has_aux argument to jacfwd and jvp

### LINKED ISSUE #392: has_aux flag for jacfwd (and maybe hessian)

vfdev-5:
Description:
- Added has_aux argument to jacfwd and jvp
- Updated tests and docs

Fixes #392


zou3519:
nit: "Currently, #404 is merged into this branch" is not the case anymore, it can be removed from the PR description

functorch/_src/eager_transforms.py
```diff
@@ -314,7 +314,7 @@ def jacrev(func: Callable, argnums: Union[int, Tuple[int]] = 0, *, has_aux=False
         returns the Jacobian of :attr:`func` with respect to the arg(s) at
         :attr:`argnums`. If ``has_aux is True``, then the returned function
         instead returns a ``(jacobian, aux)`` tuple where ``jacobian``
-        is the Jacobian and ``aux`` is auxiliary objects returned by ``func``.
+        is the Jacobian and ``aux`` is auxiliary objects returned by :attr:`func`.
```
zou3519, line 317: Why the change? I thought :attr: is for attributes on objects -- func is an argument to the function.

functorch/_src/eager_transforms.py
```diff
@@ -670,7 +670,9 @@ def safe_unpack_dual(dual, strict):
     return primal, tangent
 
 
-def jvp(func, primals, tangents, *, strict=False):
+def jvp(
+    func: Callable, primals: Any, tangents: Any, *, strict: bool = False, has_aux: bool = False
+) -> Union[Tuple[Any, Any], Tuple[Any, Any, Any]]:
```
zou3519, line 675: I'm not sure that having the type of the return is helpful to users, what do you think?

functorch/_src/eager_transforms.py
```diff
@@ -754,7 +771,13 @@ def jvp(func, primals, tangents, *, strict=False):
                 partial(_undo_create_differentiable, level=level), primals_out)
             tangents_out = tree_map(
                 partial(_undo_create_differentiable, level=level), tangents_out)
-            return tree_unflatten(primals_out, spec), tree_unflatten(tangents_out, spec)
+
+            primals_out_unf = tree_unflatten(primals_out, spec)
+            tangents_out_unf = tree_unflatten(tangents_out, spec)
```
zou3519, line 776: nit: primals_out_unflatten is more verbose but it's easier to understand than `unf`

functorch/_src/eager_transforms.py
```diff
@@ -745,6 +753,15 @@ def jvp(func, primals, tangents, *, strict=False):
...
             duals = tree_unflatten(flat_duals, primals_spec)
             result_duals = func(*duals)
+            if has_aux:
+                if not (isinstance(result_duals, tuple) and len(result_duals) == 2):
+                    raise RuntimeError(
+                        f"{jvp_str}: output of function f should be a tuple: (output, aux) "
+                        "if has_aux is True"
+                    )
+                result_duals, aux = result_duals
+                aux = _undo_create_differentiable(aux, level)
```
zou3519, line 763: _undo_create_differentiable only works on Tensors right? Does jax permit returning arbitrary aux output? (If so, we don't necessarily have to fix that here, I'm just trying to understand)

functorch/_src/eager_transforms.py
```diff
@@ -854,11 +884,26 @@ def wrapper_fn(*args):
...
             return jvp_out
 
         results = vmap(push_jvp)(basis)
+        if has_aux:
+            results, aux = results
+            # aux is in the standard basis format, e.g. NxN matrix
+            # We need to fetch the first element as original `func` output
+            flat_aux, aux_spec = tree_flatten(aux)
+            flat_aux = [value[0] for value in flat_aux]
+            aux = tree_unflatten(flat_aux, aux_spec)
```
zou3519, line 901: Hmm, I don't follow -- what's going on here? Do you have an example?

zou3519: Oh, I get it now... we didn't actually want to vmap over the aux computation, we want to return the actual aux output of f(*primals) to the user...

test/test_eager_transforms.py
```diff
@@ -1255,18 +1255,10 @@ def f(x):
...
-            self.assertEqual(aux, expected_aux)
-        else:
-            result, aux = jacapi(f)(x)
-            self.assertEqual(result, torch.eye(3, 3, device=device))
-            expected_aux = {'a': -torch.diag(x.sin()), 'b': [torch.diag(1.0 + x.tan() ** 2)]}
-            self.assertEqual(aux, expected_aux)
+        result, aux = jacapi(f, has_aux=True)(x)
+        self.assertEqual(result, torch.eye(3, 3, device=device))
+        expected_aux = {'a': x.cos(), 'b': [x.tan()]}
+        self.assertEqual(aux, expected_aux)
```
zou3519, line 1261: Could we also add a test for jacapi that checks the error cases? E.g. if aux is a number, then the current state is that we error out. It'll be good to also check the error message for that is nice.

test/test_eager_transforms.py
```diff
@@ -1782,6 +1774,44 @@ def composite_output(x):
...
+
+        x = torch.randn(3, device=device)
+        t = torch.randn(3, device=device)
+
+        out, jvp_out, aux = jvp(f, (x, ), (t, ), has_aux=True)
+        expected_out, expected_aux = f(x)
+        self.assertEqual(out, expected_out)
+        self.assertEqual(aux, expected_aux)
+        self.assertEqual(jvp_out, t * x.cos())
+
```
zou3519, line 1814: Sorry this is my fault because I didn't write the tests for these, but since we are here... could you also add a test for vjp that tests the error case (that it errors out when aux is not a pytree-of-tensors) and tests for grad with has_aux = True? I'm also happy to handle this as a followup if you don't want to.

zou3519 (commented):
The code for this looks great! I had some comments on beefing up the testing (whenever someone adds a user-facing feature we generally ask for a lot more tests to cover all the cases, compared to something like a batching rule where the tests get auto-generated)

functorch/_src/eager_transforms.py
```diff
@@ -139,7 +139,7 @@ def _autograd_grad(outputs, inputs, grad_outputs=None, retain_graph=False, creat
 
 
 # How do we increment and decrement the nesting? I don't think we can.
-def vjp(func: Callable, *primals, has_aux=False):
+def vjp(func: Callable, *primals, has_aux: bool = False) -> Callable:
```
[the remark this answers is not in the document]
vfdev-5: Oh, you are right, this is wrong as return type. Thanks for catching that.

functorch/_src/eager_transforms.py
vfdev-5: > Why the change?

To make the docs a bit more uniform. It says 
```
        Returns a function that takes in the same inputs as :attr:`func` and
        returns the Jacobian of :attr:`func` with respect to the arg(s) at
        :attr:`argnums`. If ``has_aux is True``, then the returned function
        instead returns a ``(jacobian, aux)`` tuple where ``jacobian``
        is the Jacobian and ``aux`` is auxiliary objects returned by :attr:`func`.
```

So, the last `:attr:'func'` is the same as the one from 
```
        Returns a function that takes in the same inputs as :attr:`func` and
        returns the Jacobian of :attr:`func` with respect to the arg(s) at
```

functorch/_src/eager_transforms.py
vfdev-5: I can remove it as we do not run any mypy checks. I added it as added the types for input signature.

functorch/_src/eager_transforms.py
zou3519: In general I think we are using :attr: wrong, but it doesn't have any negative effects so this is fine :)

vfdev-5: Yes, I agree that in rendered docs :attr:`func` does not do what I was thinking. Anyway, there is a room of improvements for the docstrings as well.

functorch/_src/eager_transforms.py
vfdev-5: I made another PR where trying to remove restrictions on aux. Here I'm just keeping similar approach as other ops doing

zou3519 (approved):
Awesome, thank you!

zou3519:
So the reason why I haven't merged this yet is that our CI looks very sad, I'm working on fixing that right now

zou3519:
Test failures look real?

zou3519 (approved):

PR #420, opened 2022-01-27
normal support

samdow:
* #523
* #521
* __->__ #420
* #511


samdow:
TODO: This is a relatively big PR, so I'll break it up into normal.float_float, normal.Tensor_float, and normal.Tensor_Tensor

test/test_vmap.py
```diff
@@ -3530,6 +3529,7 @@ def test_distributions(self, device, randomness, use_generator, batched_input):
         supported_ops = [
             lambda _, t: torch.normal(t, 1., **kwargs),
             lambda _, t: torch.normal(0., torch.abs(t), **kwargs),
+            lambda _, t: torch.normal(t, torch.abs(t), **kwargs),
```
zou3519, line 3532: torch.normal(a, b) 

For Tensors a and b, there are 12 cases to test:
- {error, same, different} x {a batched, b batched, a and b batched, nothing is batched}

I think we're only testing {error, same, different} x {a and b batched}?

functorch/csrc/BatchRulesHelper.cpp
```diff
@@ -83,6 +83,24 @@ Tensor maybePadToLogicalRank(const Tensor& tensor, optional<int64_t> has_bdim, i
   return tensor.view(new_sizes);
 }
 
+void check_randomness(RandomnessType randomness, bool any_tensor_batched) {
+  TORCH_CHECK(
+    randomness != RandomnessType::Error,
+    "vmap: called random operation while in randomness error mode. Please either use the "
```
zou3519, line 89: nit: would be nice if the name of the random operation was included in the error message

functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -89,6 +89,48 @@ struct BinaryPointwiseBatchRuleHelper<F, Func, typelist<T1, T2, T...>> {
...
+    optional<int64_t> other_bdim;
+    std::tie(other_value, other_bdim) = unwrapTensorAtLevel(other, cur_level);
+
+    check_randomness(randomness, (tensor_bdim || other_bdim));
+    if (randomness == RandomnessType::Different && !tensor_bdim && !other_bdim) {
+      auto shape = tensor_value.sizes();
+      VmapDimVector shapeVec(shape.begin(), shape.end());
+      shapeVec.insert(shapeVec.begin(), maybe_layer->batchSize());
+      tensor_value = tensor_value.unsqueeze(0).expand(shapeVec);
+      tensor_bdim = 0;
```
zou3519, line 117: Do we even need this case? Does `auto res = _binary_pointwise_batch_rule<F, Func, T...>(` handle this?

It looks like what we're doing here is going to be faster than `auto res = _binary_pointwise_batch_rule<F, Func, T...>(` so it's probably good to keep this around, but just to check my understanding

zou3519 (commented):
code looks correct, some questions

functorch/csrc/BatchRulesBinaryOps.cpp
samdow: Just in case I misunderstood, putting both special cases here:
For the different, neither element is batched case: I believe that `_binary_pointwise_batch_rule` assumes that at least one of the tensors is batched (i.e. it never calls like `ensure_has_batch_dimension` or has a check that at least one bdim exists). So this basically just chooses the first tensor and expands that to make it batched. It probably could be done with `ensure_has_batch_dimension` though just to use those helper functions

For the same case: we could, we would just need some special handling again since `_binary_pointwise_batch_rule` always returns a bdim of 0 (it assumes the outcome must be batched). This does avoid all the checks + calls to helper functions (even if they won't do anything in this case)

- [ ] Add a couple comments explaining the reasoning for each case (speed, batch rule doesn't add a batch dim) @samdow

PR #423, opened 2022-01-28
Enabled any type of aux output

vfdev-5:
Description:
- Enabled any type of aux output (previously, aux could be only pytree tensors due to `aux = _undo_create_differentiable(aux, level)`)
- Parity with JAX API:
```python
import jax.numpy as jnp
import jax

def f(x):
    return (x * 2).sum(), "abc"

x = jnp.array([1., 1.5])
out = jax.grad(f, has_aux=True)(x)
>
(DeviceArray([2., 2.], dtype=float32), 'abc')
```

### DIFF
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -33,7 +33,7 @@ def _create_differentiable(inps, level=None):
     return tree_map(create_differentiable, inps)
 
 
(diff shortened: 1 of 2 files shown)


Chillee:
Makes sense to me.

Chillee (approved):

PR #425, opened 2022-01-31
Fix CI

zou3519:
- Update to new FuncTorchTLS interface from pytorch/pytorch


zou3519:
Shipping for now, we've got tests to fix.

Issue #393, opened 2022-01-11
vmapvjp failure on __getitem__


zou3519:
Probably requires a change to index.Tensor's backward formula in pytorch/pytorch

https://github.com/pytorch/functorch/blob/48eec0a05bb169cd9e4b699dc191670d47ee61ca/test/test_ops.py#L560

PR #426, opened 2022-01-31
index_put : batch-rule

kshitij12345:
Adds batching rule for `index_put`

This also fixes: #393

### DIFF
--- a/functorch/csrc/BatchRulesScatterOps.cpp
+++ b/functorch/csrc/BatchRulesScatterOps.cpp
@@ -9,6 +9,9 @@
 #include <ATen/Operators.h>
 #include <functorch/csrc/PlumbingHelper.h>
 #include <functorch/csrc/BatchedFallback.h>
+#include <ATen/native/TensorAdvancedIndexing.h>
+#include <ATen/native/IndexKernel.h>
+#include <ATen/native/IndexingUtils.h>
 
 
 namespace at { namespace functorch {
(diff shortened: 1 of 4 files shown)


zou3519:
@kshitij12345 can you rebase this please? The CI looks like it's stuck at a build failure we fixed yesterday

functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -220,6 +224,33 @@ namespace {
 
     return std::make_tuple(self_, indices_, values_);
   }
+
+  auto index_put_plumbing_helper(const Tensor &self,
+                                 const List<optional<Tensor>> &indices,
+                                 const Tensor &values, int64_t cur_level) {
```
zou3519, line 230: nit: "unpackSelfAndIndicesAndValuesAtCurrentLevel" is many more characters but definitely more descriptive. It took me a couple of seconds before I understood what was going on here because we have a lot of functions named similarly: index_put_plumbing_helper, index_put_batch_rule_helper

functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -179,16 +179,20 @@ namespace {
...
-    const auto batch_size = self_.size(0);
+    // for inplace variants `index_put_` and `_index_put_impl_` we find the batch_size
+    // here while for `index_put` does it outside of this function.
+    const auto batch_size = opt_batch_size ? opt_batch_size.value() : self_.size(0);
+    self_ = ensure_has_bdim(self_, self_bdim.has_value(), batch_size);
     values_ = ensure_has_bdim(values_, values_bdim.has_value(), batch_size);
     TORCH_INTERNAL_ASSERT(indices.size() == indices_bdims.size());
-    std::vector<optional<Tensor>> indices_ = batchIndices(indices, indices_bdims, self_.size(0), self_bdim, values_bdim);
+
+    // we make sure that `self` has bdim at `0`.
```
zou3519, line 194: nit: "we've already made sure that `self` has bdim at `0`" to clarify that we've already done this (the way this is currently written is ambiguous: did we already do this, or does this happen in batchIndices?)

test/functorch_additional_op_db.py
```diff
@@ -342,3 +341,36 @@ def sample_inputs_binary_cross_entropy(op_info, device, dtype, requires_grad, lo
         supports_out=False,
         supports_forward_ad=True,
     ))
+
+
+def sample_inputs_index_put(op_info, device, dtype, requires_grad, **kwargs):
+    make_arg = partial(make_tensor, dtype=dtype, device=device, requires_grad=requires_grad)
+    make_idx = partial(make_tensor, dtype=torch.long, device=device, requires_grad=False)
+    S = 5
```
zou3519, line 349: index_put is a pretty core operation in PyTorch. Could we add some more tests? Maybe a manual test like https://github.com/pytorch/functorch/blob/ec69443bdc1c5ee6b1e143383dc20bb392776597/test/test_vmap.py#L3353-L3364 is sufficient

zou3519 (commented):
I had some really minor nits and then a suggestion to add some more test cases for index_put. The OpInfo for index_put only has 3(?) test cases which doesn't seem like a lot --index_put is a pretty core pytorch operation.

test/functorch_additional_op_db.py
kshitij12345: I think the reason for having manual tests for the in place variant is because we want `self` to always have a bdim. With out-of-place `index_put`, we can use `get_fallback_and_vmap_exhaustive` to generate variety of batched inputs. So will try to add more cases in the OpInfo sample.

zou3519 (approved):
LGTM, thank you!

PR #441, opened 2022-02-03
Added Grad context to AOTAutograd

### LINKED ISSUE #442: `nnc_jit` fails with inner autograd call
```
import torch
from functorch.compile import nnc_jit


def foo(x):
    y = torch.exp(x)
    # no gradient information in y...
    print(y)
    z = torch.autograd.grad(y, x)
    return z

print("Before:")
x = torch.tensor(1.0, requires_grad=True)
print(foo(x))

print("After NNC JIT:")
x = torch.tensor(1.0, requires_grad=True)
print(nnc_jit(foo)(x))
```

```
# RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

Chillee:
@anijain2305 Fixes #442

### DIFF
--- a/functorch/_src/aot_autograd.py
+++ b/functorch/_src/aot_autograd.py
@@ -88,7 +88,7 @@ def _reshape_alias(x, shape, strides):
     return aten.view(x, shape)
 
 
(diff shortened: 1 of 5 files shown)


anijain2305 (approved):
LGTM

PR #447, opened 2022-02-03
Make min-cut rematerialization optional.

hyunwoongko:
Some users may not want to use min-cut rematerialization for something like activation checkpointing. 
So I changed this to optional. Does this make sense?

cc @Chillee


hyunwoongko:
Some tests are failed. but I think these are not related with my PR.

PR #452, opened 2022-02-04
Fix tests

zou3519:


test/test_ops.py
```diff
@@ -544,7 +539,6 @@ def vjp_of_vjp(*args_and_cotangents):
         skip('linalg.svdvals'),
```
Lezcano, line 539: missed this one I believe @zou3519

PR #453, opened 2022-02-04
Switched docs theme to pytorch-theme

vfdev-5:
Deployement to check the rendering:
- https://vfdev-5.github.io/functorch/


zou3519:
@vfdev-5 sorry, looking at the rendering reminded me of the actual reason why I didn't use the pytorch theme. The top bar links to various things that are unrelated to functorch!

For examples, tutorials links to PyTorch tutorials, not functorch tutorials...

vfdev-5:
Ok, I can customize that quickly. Do you want to change the topleft logo by something "functorch" ?

zou3519:
Does the theme look OK if we delete the entire top bar? (and the logo?)

Also, there are things missing, like the + in the below image:
image
image

vfdev-5:
I won't remove the header just replace titles and links. Like in pytorch.org/ignite

As for toctree and list expansion I can check that. 
On the other hand I do not insist on switch, if you prefer rtd theme, let's just refresh it

zou3519:
> I won't remove the header just replace titles and links. Like in pytorch.org/ignite

That SGTM, I didn't realize it was easy to replace those.

> As for toctree and list expansion I can check that. On the other hand I do not insist on switch, if you prefer rtd theme, let's just refresh it

I think it'll be good to be consistent with the PyTorch theme, especially because eventually we do want to make sure our docs render nicely under the PyTorch theme. So let's try to make the pytorch theme work for us if it is not too much work.

> Ok, I can customize that quickly. Do you want to change the topleft logo by something "functorch" ?

That would be nice, if we had a graphic designer lol.

docs/source/functorch.rst
```diff
@@ -11,7 +11,9 @@ Function Transforms
 
     grad
     grad_and_value
+    hessian
     jacrev
+    jacfwd
     vmap
     vjp
```
zou3519, line 18: Add jvp too please?

docs/source/conf.py
```diff
@@ -157,44 +154,22 @@
...
-    'css_files': [
-        'https://fonts.googleapis.com/css?family=Lato',
-        '_static/css/pytorch_theme.css'
-    ],
+    "collapse_navigation": False,
+    "display_version": True,
+    "logo_only": True,
+    "pytorch_project": "docs",
+    "navigation_with_keys": True,
+    "analytics_id": "UA-117752657-2",
```
zou3519, line 170: Is this PyTorch's google analytics ID?

zou3519 (approved):
One really minor comment. Also, it's really difficult to tell that this is functorch documentation because the logo says "PyTorch", so at some point (doesn't need to be in this PR) we should try to replace it with text that says functorch or something

docs/source/conf.py
vfdev-5: Yes, it is the same as in vision: 

https://github.com/pytorch/vision/blob/dad6e6a44d63272459b54175c92ed22955eb4207/docs/source/conf.py#L126

and pytorch:

https://github.com/pytorch/pytorch/blob/b9d4c0c7a9689982feb949f9609cec143c4a7268/docs/source/conf.py#L393

vfdev-5:
@zou3519 I updated the docs:
- added vjp entry
- added `functorch` text logo

docs/source/conf.py
zou3519: Awesome!

zou3519 (approved):
Thanks! This looks great

PR #463, opened 2022-02-07
Doc strings

anijain2305:
@Chillee PTAL

I used black to format as well.


anijain2305:
Horace confirmed offline that this looks good. Merging for now.

PR #466, opened 2022-02-08
Fix CI

samdow:
xfails conv_transpose3d...hopefully fixes CI finally

Once this fixes, I'll add an issue for this xfail


functorch/_src/aot_autograd.py
```diff
@@ -476,15 +476,17 @@ def _named_members(mod, get_members_fn, prefix='', recurse=True, remove_duplicat
 
 
 def _named_parameters(mod, prefix: str = '', recurse: bool = True, remove_duplicate: bool = True):
-    gen = mod._named_members(
+    gen = _named_members(
```
samdow, line 479: cc @Chillee @anijain2305 JFYI that I added this change so that it calls the functions here instead of the member functions that don't have these params yet (and were giving CI errors). I wasn't sure if I should add an extra comment here but wanted to flag for you two

PR #471, opened 2022-02-09
Revert "Enabled any type of aux output"

vfdev-5:
Reverts pytorch/functorch#423

We have to more about testing #423 and the implications it could have. Let's keep aux output as tensor for now


zou3519 (approved):
Thank you!

PR #494, opened 2022-02-15
functorch manylinux builds

zou3519:


zou3519:
@malfet -- this is already merged, but I would appreciate a read through b/c I'm not sure what I'm doing :)

.github/workflows/wheels.yml
```diff
@@ -8,29 +8,54 @@ on:
...
+    runs-on: macos-latest
+    strategy:
+      matrix:
+        python_version: [ "3.7", "3.8", "3.9" ]
     steps:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: ${{ matrix.py_ver }}
+          python-version: ${{ matrix.python_version }}
```
zou3519, line 45: @malfet -- the comment here https://github.com/actions/setup-python/issues/26 is a bit troubling. Ideally our binaries would be compatible with the 10.9 SDK (like the rest of PyTorch), but the thread suggests that github actions sets up Python to be compatible with the 10.14 SDK. Does this mean we have to install python a different way? (e.g. conda like how torchvision and torchaudio are doing it).

PR #504, opened 2022-02-16
Add colab runnable version of jacobians_hessians tutorial, and 'run with colab' badge

lessw2020:
This PR adds a colab runnable tutorial for the current 'jacobians_hessians' notebook.
There are 2 main updates:
1 - minor -The notebook jacobians_hessians has a 'run with colab' link added to the top. This link opens the second file, jacobians_hessians_colab.ipynb directly in colab.

2 - major - The main change is the addition of jacobians_hessians_colab.ipynb which integrates the install and building of functorch in colab, adds a restart function and then lets the user directly run the functorch tutorial. 
There are a few editing additions such as adding a relative performance function to highlight the relative perf improvements of certain operations, and a few other additions but largely stays true to the notebook tutorial. 

Tutorial in colab:
colab_functorch_update

Emphasis on perf improvement (highlight not part of colab lol) :
functorch_jacrev_perf


notebooks/jacobians_hessians.ipynb
```diff
@@ -7,12 +7,22 @@
    "source": [
     "# Jacobians, hessians, and more: composing functorch transforms\n",
     "\n",
+    "<a href=\"https://colab.research.google.com/github/pytorch/functorch/blob/main/notebooks/jacobians_hessians_colab.ipynb\">\n",
+    "  <img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/>\n",
+    "</a>\n",
+    "\n",
```
zou3519, line 13: Question: are jacobian_hessians.ipynb and jacobians_hessians_colab.ipynb copies of each other, or is there some mechanism that deduplicates code between them?

It might not be the best experience if to update a tutorial someone has to update both jacobian_hessians.ipynb and jacobians_hessians_colab.ipynb

lessw2020: Currently the code is similar but has differences (example the get perf to compare benchmark, the collaboration restart function, a few clearer variable names).
Thus no easy code dedupe ATM. Agree that would be easier, flip side is updates to these might be few relatively and may need to keep things colab specific regardless.example dim for hessian example is large by a100 in main, vs reduced that for colab or it crashes colab kernel).

notebooks/jacobians_hessians_colab.ipynb
```diff
@@ -0,0 +1,1135 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
```
zou3519, line 3: So, right now, our docs build is set up to include all the notebooks under notebooks/* in the documentation website: https://pytorch.org/functorch/main/tutorials.html. We should probably make sure the jacobians_hessians_colab does not get included as a tutorial

notebooks/jacobians_hessians.ipynb
zou3519: Let me think about if there's anything easy we can do to dedupe, if not, this seems fine (and I'll file an issue to fix this in the future).

The main reason why we used a100 was because that's what is on the machine I have :P, I wouldn't mind dropping that benchmark number to keep these two notebooks in sync

zou3519 (commented):
This looks pretty good! I had two comments/concerns. Let me think a little more on if there's anything easy we can do to dedup, otherwise this should be fine

zou3519 (approved):
I have some ideas for how to deduplicate this (we are probably going to swap out nbsphinx for myst-nb (it has some nicer properties compared to nbsphinx) and then put the install instructions in a collapsible section). That being said, it could be tricky and it's beyond the scope of the bootcamp task so I'll file a followup for it :).

Otherwise, this LGTM -- but please move the jacobian_hessian_colab.ipynb* files to a colab/ directory under functorch in a followup

PR #505, opened 2022-02-17
Add Batch Norm module utilities to not track running stats

samdow:
We want to let users run the maml_omniglot example as it was originally written, using batch norm. Given how vmap works though, this can't be done with batch norm that tracks running statistics because those are updated in place. That's fine, the example shouldn't be using the running statistics anyway.

The issues with batch norm were determined to be numerical precision issues around 0. The example in the examples folder uses batch norm and still trains, suggesting it's an error around the value being near 0:
maml-accs

So this:
 - switches the maml_omniglot example in the examples folder to use Batch Norm
 - adds utilities for switching batch norm on prebuilt models. This isn't needed for maml_omniglot because the net in that example is built inline. However it would be useful for examples like torchvision models
 - tests the newly added utilities to run without tracking stats

In a separate PR I will:
- update the batch norm error message to be more descriptive
- setup a page describing the ways to make batch norm work with vmap

### DIFF
--- a/test/test_eager_transforms.py
+++ b/test/test_eager_transforms.py
@@ -15,15 +15,14 @@ import unittest
 import warnings
 import math
 from torch.testing._internal.common_device_type import instantiate_device_type_tests, onlyCPU
-from functorch.experimental.batch_norm_replacement import replace_all_batch_norm_modules
 from torch.testing._internal.common_dtype import get_all_fp_dtypes
 from functools import partial
-from functorch.experimental import copy_and_replace_all_batch_norm_modules, replace_all_batch_norm_modules
+from functorch.experimental import replace_all_batch_norm_modules_
 
 import functorch
 from functorch import (
-    grad, vjp, vmap, jacrev, jacfwd, grad_and_value,
-    make_functional, make_functional_with_buffers,
+    grad, vjp, vmap, jacrev, jacfwd, grad_and_value, hessian,
+    jvp, make_functional, make_functional_with_buffers,
 )
 from functorch._src.make_functional import (
     functional_init, functional_init_with_buffers,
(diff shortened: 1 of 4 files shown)


test/test_eager_transforms.py
```diff
@@ -2057,6 +2052,112 @@ def compute_loss(new_params, buffers, x, y):
...
+        losses = [compute_loss(x_spt[i], y_spt[i], x_qry[i], y_qry[i])[0]
+                  for i in range(num_tasks)]
+        expected_grads = torch.autograd.grad(sum(losses), params)
+
+        self.assertEqual(result_grads[2:], expected_grads[2:])  # should pass
+        self.assertEqual(result_grads[:2], expected_grads[:2])  # should fail
+
+    @parametrize('copy', [True, False])
+    @parametrize('originally_track_running_stats', [True, False])
+    def test_update_batch_norm(self, device, copy, originally_track_running_stats):
```
samdow, line 2119: This shows that `vmap(grad` works on batch norm, for both cpu and cuda. It's just the math in omniglot that messes it up

test/test_eager_transforms.py
```diff
@@ -2057,6 +2052,112 @@ def compute_loss(new_params, buffers, x, y):
...
+        # TODO: there appears to be precision issues for float32
+        dtype = torch.double
+
+        n_way = 5
+        n_inner_iter = 2
+        num_tasks = 2
+
+        net = nn.Sequential(
+            nn.Conv2d(1, 64, 3),
+            nn.BatchNorm2d(64, momentum=1, affine=True, track_running_stats=False),
```
samdow, line 2065: This works if it's GroupNorm but fails if it's BatchNorm2d or LayerNorm, both on cpu and cuda

PR #506, opened 2022-02-17
Colab ready version, per sample gradients tutorial

lessw2020:
adds:
colab edition of per sample gradients with functorch build and install.
colab badge to current docs/notebook tutorial to link the above.


zou3519 (commented):
Did you mean to include the files from jacobians_hessians in this PR?

lessw2020:
> Did you mean to include the files from jacobians_hessians in this PR?

It was not intentional - an artifact of thinking these would be done in serial, so I hadn't originally branched from my fork.  
But since (yesterday) the bonus files were identical to the earlier PR, talked to Horace and he said not to worry about it.
That said, will branch in the future to avoid this!

PR #522, opened 2022-02-18
Made codegen 1:1 instead of many:1

zou3519:


codegen/codegen_outofplacebatching.py
```diff
@@ -247,17 +188,8 @@ def __init__(self, schema, ops):
...
-        if unique_id in unique_schemas:
-            unique_schemas[unique_id].ops.append(op)
-        else:
-            unique_schemas[unique_id] = Container(schema, [op])
-
-    codes = [lower(*container.schema, i, container.ops)
-             for i, container in enumerate(unique_schemas.values())]
+    codes = [gen_plumbing(*schema) for schema in schemas]
+    print("#pragma once")
     print("#include <functorch/csrc/OutOfPlacePlumbing.h>")
```
vfdev-5, line 193: @zou3519 shouldn't we remove this line as generated file (functorch/csrc/OutOfPlacePlumbing.h) includes itself ?

zou3519: uh oh, good catch

vfdev-5: pragma once should help to avoid inf loop

PR #527, opened 2022-02-22
Quick fix for torch.trace backward

### LINKED ISSUE #501: vmap fails for hessian of torch.trace
Hello, 

I've been trying to vmap the hessian of torch.trace and it fails, however, calculating the jacobian is just fine. Here's a reproducible error,
```
import torch
from functorch import jacrev, vmap

matrix = torch.randn(2,4,4)
deriv1 = vmap(jacrev(torch.trace), in_dims=(0))(matrix)
print(deriv1)

deriv2 = vmap(jacrev(jacrev(torch.trace)), in_dims=(0))(matrix)
print(deriv2)
```

which results in  the following error,
```
RuntimeError: vmap: aten::index_fill_(self, *extra_args) is not possible because there exists a Tensor `other` in extra_args that has more elements than `self`. This happened due to `other` being vmapped over but `self` not being vmapped over at level 2. Please try to use out-of-place operators instead of aten::index_fill_. If said operator is being called inside the PyTorch framework, please file a bug report instead.
```

I assume this error just because pytorch calls an in-place error and it's not a fault with functorch's `jacrev` function? I only ask this because this works for `jacrev(jacfwd(torch.trace))`.

For example, 
```
import torch
from functorch import jacfwd, jacrev, vmap

matrix = torch.randn(2,4,4)
try:
    deriv1 = vmap(jacrev(torch.trace), in_dims=(0))(matrix)
    print("jacrev(torch.trace) passed")
except:
    print("jacrev(torch.trace) failed")

try:
    deriv2 = vmap(jacrev(jacrev(torch.trace)), in_dims=(0))(matrix)
    print("jacrev(jacrev(torch.trace)) passed")
except:
    print("jacrev(jacrev(torch.trace)) failed")

try:
    deriv3 = vmap(jacrev(jacfwd(torch.trace)), in_dims=(0))(matrix)
    print("jacrev(jacfwd(torch.trace)) passed")
except:
    print("jacrev(jacfwd(torch.trace)) failed")
```
returns,
```
jacrev(torch.trace) passed
jacrev(jacrev(torch.trace)) failed
jacrev(jacfwd(torch.trace)) passed
```
where the error is the same RunTimeError stated above! 

Thank you! :)

vfdev-5:
Description:

- Quick fix for torch.trace backward
- Updated tests

Note:

We may want to use index_put_ instead of index_put

Fixes #501

### DIFF
--- a/functorch/csrc/BatchRulesViews.cpp
+++ b/functorch/csrc/BatchRulesViews.cpp
@@ -155,6 +155,17 @@ Tensor trace_decomp(const Tensor& self) {
   return at::sum(at::diagonal(self));
 }
 
+Tensor trace_backward_decomp(const Tensor& grad, IntArrayRef sizes) {
+  if (sizes.size() != 2) {
+    throw std::runtime_error("expected matrix input");
+  }
+  auto grad_input = at::zeros(sizes[0] * sizes[1], grad.options());
+  auto indices = at::arange(0, grad_input.numel(), sizes[1] + 1, grad.options().dtype(at::kLong));
+  // Workaround using index_put instead of yet unsupported index_fill_
+  grad_input = grad_input.index_put({indices}, grad);
+  return grad_input.view(sizes);
+}
+
 std::tuple<Tensor,optional<int64_t>> flip_batch_rule(const Tensor& self, optional<int64_t> self_bdim, IntArrayRef dims) {
   auto self_ = moveBatchDimToFront(self, self_bdim);
   VmapDimVector new_dims;
@@ -499,6 +510,7 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   m.impl("flatten.using_ints", static_cast<decltype(&ATEN_FN2(flatten, using_ints))>(native::flatten));
   VMAP_SUPPORT(flip, flip_batch_rule);
   m.impl("trace", trace_decomp);
+  m.impl("trace_backward", trace_backward_decomp);
   VMAP_SUPPORT(tril, VARIADIC_BDIMS_BATCH_RULE(ATEN_FN(tril)));
   VMAP_SUPPORT(triu, VARIADIC_BDIMS_BATCH_RULE(ATEN_FN(triu)));
   VMAP_SUPPORT(repeat, repeat_batch_rule);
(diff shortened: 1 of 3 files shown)


functorch/csrc/BatchRulesViews.cpp
```diff
@@ -499,6 +510,7 @@ TORCH_LIBRARY_IMPL(aten, FT_BATCHED_KEY, m) {
   m.impl("flatten.using_ints", static_cast<decltype(&ATEN_FN2(flatten, using_ints))>(native::flatten));
   VMAP_SUPPORT(flip, flip_batch_rule);
   m.impl("trace", trace_decomp);
+  m.impl("trace_backward", trace_backward_decomp);
```
zou3519, line 513: I think we might need to register the decomposition over at https://github.com/pytorch/functorch/blob/main/functorch/csrc/PyTorchOperatorHacks.cpp#L192 .

Our composition tests don't actually cover the following:
```
import torch
from functorch import grad, vmap

def sum_grad_trace(x):
  return grad(torch.trace)(x).sum()

x = torch.randn(2, 2, 2)
vmap(grad(sum_grad_trace))(x)
```
does this pass under your PR?

vfdev-5: Unfortunately, no, this does not pass

test/test_vmap.py
```diff
@@ -2975,6 +2975,17 @@ def test_trace(self, device):
...
+        # TODO: put this test to where it will be appropriate
+        from functorch import grad
+
+        x = torch.randn(3, 2, 2, device=device)
+
+        def sum_grad_trace(x):
+            return grad(torch.trace)(x).sum()
+
+        output = vmap(grad(sum_grad_trace))(x)
+        self.assertEqual(output, torch.zeros_like(output))
```
vfdev-5, line 2987: @zou3519 I added temporarily your test in `test_trace` to ensure that we decompose correctly the op. Let me know if there is a more appropriate place for such check.

zou3519: I think this is fine

test/test_vmap.py
```diff
@@ -2975,6 +2975,17 @@ def test_trace(self, device):
         x = torch.randn(2, 3, device=device, requires_grad=True)
         self._batched_grad_test(Tensor.trace, (x,))
 
+        # TODO: put this test to where it will be appropriate
```
zou3519, line 2978: remove todo maybe?

zou3519 (approved):
lgtm, might want to remove the todo

PR #530, opened 2022-02-23
Added docs deployment step in GHA

### LINKED ISSUE #500: Set up main docs build
Set up a docs build in the CI. Upon changes to the main branch (or maybe on a nightly basis?) we should push to https://pytorch.org/functorch/main/ .

Motivation:
- Automation: Saves me/us some hours of remembering how to set up the docs build and needing to rebuild the docs periodically :)
- We can totally reuse this for the documentation for the release branch as well

vfdev-5:
Description:
- Added GHA "docs deploy" job based on "build-docs" job
- Using an "action" to commit and push to the branch.
  - use `git reset --soft <fixed-commit>` to squash history
  - History on gh-pages: https://github.com/vfdev-5/functorch/commits/gh-pages
- Runned GHA that pushed to gh-pages on my fork:
  - https://github.com/vfdev-5/functorch/actions/runs/1888876266

Fixes #500

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,4 +1,4 @@
-name: Build Docs
+name: Build and Deploy Docs
 on:
   pull_request:
     types: [opened, synchronize, reopened]
@@ -35,5 +35,40 @@ jobs:
       - name: Upload docs as GHA artifact
         uses: actions/upload-artifact@v2
         with:
-          name: docs.whl
+          name: built-docs
           path: docs/build/html
+
+  deploy-docs:
+    needs: [build-docs]
+    runs-on: ubuntu-latest
+    if: (github.ref == 'refs/heads/main' && github.event_name == 'push')
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+          fetch-depth: 3
+
+      - name: Download docs artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: built-docs
+          path: /tmp/docs
+
+      - name: Copy built docs to nightly
+        run: |
+          cp -R /tmp/docs/* nightly/
+          # Manipulations to squash the history:
+          # 1) reset to a fixed commit
+          # 2) use github-actions-x/commit to commit everything as a single commit
+          git reset --soft 218f32faf8b8cc5236bc49ea77ef7d222e21e5d6
+
+      # Can not use commit --amend to squash history
+      - name: Commit and push to gh-pages
+        uses: github-actions-x/commit@v2.8
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          push-branch: 'gh-pages'
+          commit-message: 'New docs'
+          force-push: 'true'
+          name: gha
+          email: gha@email.org
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -55,20 +55,28 @@ jobs:
           path: /tmp/docs
 
       - name: Copy built docs to nightly
+        id: copy-docs
         run: |
           cp -R /tmp/docs/* nightly/
-          # Manipulations to squash the history:
-          # 1) reset to a fixed commit
-          # 2) use github-actions-x/commit to commit everything as a single commit
-          git reset --soft 218f32faf8b8cc5236bc49ea77ef7d222e21e5d6
+          git log -3
+          # Set commit name and hash as variables: commit_name, commit_hash
+          echo "::set-output name=commit_name::$(git log -1 --format='%s')"
+          echo "::set-output name=commit_hash::$(git log -1 --format='%h')"
 
-      # Can not use commit --amend to squash history
+      - name: Git reset to commit/amend
+        if: ${{ steps.copy-docs.outputs.commit_name == 'auto-generated commit' }}
+        run: |
+          # if commit_name is "auto-generated commit"
+          # then go back in commit history to commit to the same commit
+          git reset --soft ${{ steps.copy-docs.outputs.commit_hash }}~1
+          git log -3
+
       - name: Commit and push to gh-pages
         uses: github-actions-x/commit@v2.8
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -35,5 +35,39 @@ jobs:
...
+          name: built-docs
+          path: /tmp/docs
+
+      - name: Copy built docs to nightly
+        run: |
+          cp -R /tmp/docs/* nightly/
+          # Manipulations to squash the history:
+          # 1) reset to a fixed commit
+          # 2) use github-actions-x/commit to commit everything as a single commit
+          git reset --soft 218f32faf8b8cc5236bc49ea77ef7d222e21e5d6
```
zou3519, line 63: What is this commit?

vfdev-5: https://github.com/pytorch/functorch/commit/218f32faf8b8cc5236bc49ea77ef7d222e21e5d6

zou3519 (approved):
This LGTM. I want to wait until the website is a little more stable before merging this

.github/workflows/docs.yml
zou3519: If nothing changes on the docs between to commits to pytorch/pytorch, is the commit essentially empty or is there something in it?

If it is essentially empty, then I'm thinking we should comment this out for now and then turn it on later -- I want to be able to manually submit changes to the ghpages branch for the next couple of weeks.

vfdev-5: This actions is triggered on every commit regardless python docstring or docs/source is unchanged (to make this action aware of specific files changes we can specify that explicitly.)
I think it could be difficult to check at runtime if commit changes or not the docs.
We may opt to another solution using another sub-action: https://github.com/devops-infra/action-commit-push which should allow `--amend` commit option. So, it means when this actions is run and rebuild the docs it would commit ammend the results on gh-pages to the last commit. What do you think ?

zou3519: Hmm, my question is, is the commit "empty" if nothing changes? Does the docs build generate different documentation based on the time of day or does it always generate the same thing if nothing has changed?

I'm trying to evaluate how large of a problem the history will become -- a lot of commits that have very little changes in them (due to us not updating the website) means we don't need to worry about having a lot of commits.

zou3519, line 63: I thought about it some more, and I think it would be better if we didn't reset the history for now (but turned it on later). The reasoning is:
1. We might do manual commits to the docs from time-to-time
2. We don't actually produce that much docs, so it'll take a while (a couple of months?) before we actually run into memory problems

What are your thoughts?

zou3519: Hmm, I like the proposal. Ideally I want some way to be able to roll the website back to a "last known good configuration" if we accidentally merge a bad change and are unable to fix it quickly.

One possible solution for that is: every week, on Sunday, have github actions do a manual commit to the gh-pages branch. The net effect of this is that we get roughly one new docs commit per week, and if something goes bad during the week, then we can roll the website back to its state from last week.

vfdev-5: Sounds good about cron-based GHA.

How about trying to make commit amend if previous commit was auto-generated and add a new commit otherwise ? In this case, we can safely roll back if needed and do not accumulated many auto-generated commits...
+ we can do everything cron-based (maybe daily).

zou3519: > How about trying to make commit amend if previous commit was auto-generated and add a new commit otherwise ? In this case, we can safely roll back if needed and do not accumulated many auto-generated commits... And we can do everything cron-based (maybe daily).

Okay! That sounds good. Hopefully we have not overengineered the system too much :)

vfdev-5:
@zou3519 I updated the GHA and here is how it worked in practice on my fork:

- I merged this PR to main and the action made a new commit : https://github.com/vfdev-5/functorch/commit/6e5fc0bc34fb7b83ab006d0fbb1980940129eda7 "auto-generated commit"

Screen Shot 2022-03-15 at 18 22 25

- Next I updated a file in `docs` on `main` branch (https://github.com/vfdev-5/functorch/commit/ced0d43bd10f9ef314e216af442bd46f6fb868d3) and GHA pushed to `gh-pages` another auto generated commit with commit/amend - https://github.com/vfdev-5/functorch/commit/8f5fdb6d85c0554f1bf6fe3e1bca65feaad20e6d "auto-generated commit"

Screen Shot 2022-03-15 at 18 28 59

- Then I manually updated two files directly on gh-pages: https://github.com/vfdev-5/functorch/commit/792dfa29b8316388e319d5d90eac8297a3526d5a and https://github.com/vfdev-5/functorch/commit/16592dde7a695b7a78a253522a243b6ad5899702

Screen Shot 2022-03-15 at 18 36 58

- Finally, I made another update on `main` : https://github.com/vfdev-5/functorch/commit/911447280772bd3e91ab38087d9a75fc46a96af2 and GHA pushed pushed to `gh-pages` another auto generated commit : https://github.com/vfdev-5/functorch/commit/8ead760319e3a8719fed48062158a7b7e4416473

Screen Shot 2022-03-15 at 18 48 37


So, I'd say things looks like working as expected.

zou3519:
Thank you @vfdev-5! I might not get to reviewing this today, but will definitely do it by Monday

zou3519 (approved):
LGTM, fingers crossed

vfdev-5:
Looks correct:
- https://github.com/pytorch/functorch/runs/5631772187?check_suite_focus=true
- https://github.com/pytorch/functorch/commit/7b7b5cfdfc4df8561470c24fabb4a47156ef9a96
- https://pytorch.org/functorch/nightly/

vfdev-5:
@zou3519 docs deployment is failing now when try to overwrite auto-generated commit on gh-pages : https://github.com/pytorch/functorch/runs/5663088291?check_suite_focus=true

Looks like gh-pages is protected branch and "Cannot force-push to this protected branch"

zou3519:
hmm, I thought I turned branch protection for the branch off. Let me check again

PR #536, opened 2022-02-25
Trace the backward pass assuming contiguous tensors

anijain2305:
Currently the AOT Autograd assumes that the forward outputs and corresponding input gradients have same strides, but thats not always true. In this PR, we force both the forward outputs and corresponding input gradients to be contiguous. 

This is suboptimal as we will call `contiguous()` when it might not be necessary. For now, we want to go ahead with the current solution, even though it might be suboptimal, because other solutions are somewhat tricky and require some overhead experiments. Tracked at #537


Chillee (approved):
LGTM!

Can you add a comment or make an issue discussing why this is suboptimal?

Issue #445, opened 2022-02-03
Some evaluation that functorch's make_functional + transforms works on torchbenchmark models


zou3519:
make_functional + our transforms have worked for most of our users out there. We should evaluate the combination on the torchbenchmark suite of models to see if there are any gaps.

vfdev-5:
Some recommendations by Richard to define the scope of the issue:

-    check that make_functional/make_functional_with_buffers + grad produces the same gradients as without using functorch (forward + backward). We only need to run a single forward+backward for this.
-    check that model ensembling using combine_state_for_ensemble + make_functional + vmap produces a quantity that is equal to just running the models in a for-loop (e.g. https://github.com/pytorch/functorch/blob/main/notebooks/ensembling.ipynb). A single forward should be sufficient for this since there is no grad transform involved.
-    check that it is possible to compute per-sample-gradients using make_functional, vmap, and grad. (e.g. https://github.com/pytorch/functorch/blob/main/notebooks/per_sample_grads.ipynb) A single forward+backward should be sufficient here but we should check this against computing the per-sample-gradient manually (as shown inside that notebook)

samdow:
This will be great! Thanks for doing this! One thing to note for the per-sample-gradient examples that I found is that if we're comparing the functorch version to a for loop version, we may want to set the loss function to have a reduction of sum (usually by default it's mean). For the example linked, in the first block of code this will look like
```python
def loss_fn(predictions, targets):
    return F.nll_loss(predictions, targets, reduction='sum')
```

The reason for this is that when doing a mean, calling the model with a single element will change the gradients because the loss is scaled by the sum of all the elements, which will be different if we're making multiple calls in a for loop

vfdev-5:
Point (1) checks are done and reported here: https://gist.github.com/vfdev-5/951e46d48edf522400b69ae594715cbc
On CPU we have gradients mismatch for certain models that have to investigate the problem.

vfdev-5:
Point (2) checks for classification/segmentation/detection torchvision models are done and reported here: 

Dropout and similar technics wont work with vmap:
```
  File "/usr/local/lib/python3.8/dist-packages/torchvision/ops/stochastic_depth.py", line 41, in stochastic_depth
    noise = noise.bernoulli_(survival_rate)
RuntimeError: vmap: We do not yet support calling random operations inside of vmap. Please perform random operations outside of vmap as a workaround
```

Other warnings:
```
[...]

PR #544, opened 2022-02-28
Added temporary odict pytree registration

vfdev-5:
Description:

- Added temporary odict pytree registration
  - combine_state_for_ensemble can work on torchvision segmentation models

Related to https://github.com/pytorch/functorch/issues/445#issuecomment-1050946177

Logs : 
```
python -u check_combine_state_for_ensemble_on_tv_segm_models.py 

Torch: 1.12.0.dev20220223+cu111
torchvision: 0.13.0.dev20220223+cu111
Functorch: 0.2.0a0+c8b640c

-- Check fcn_resnet50 model
/usr/local/lib/python3.8/dist-packages/torch/nn/functional.py:1279: UserWarning: There is a performance drop because we have not yet implemented the batching rule for aten::dropout. Please file us an issue on GitHub so that we can prioritize its implementation. (Triggered internally at  /ft/functorch/csrc/BatchedFallback.cpp:83.)
  return _VF.dropout_(input, p, training) if inplace else _VF.dropout(input, p, training)
-- Check fcn_resnet101 model
Output does not match reference: -237.7958221435547 vs -237.7958984375
-- Check deeplabv3_resnet50 model
-- Check deeplabv3_resnet101 model
Output does not match reference: -285.8777770996094 vs -285.8778076171875
-- Check deeplabv3_mobilenet_v3_large model
-- Check lraspp_mobilenet_v3_large model
```

### DIFF
--- a/functorch/_src/vmap.py
+++ b/functorch/_src/vmap.py
@@ -6,9 +6,10 @@
 
 import torch
 import functools
+from collections import OrderedDict
 from torch import Tensor
 from typing import Any, Callable, Optional, Tuple, Union, List
-from torch.utils._pytree import tree_flatten, tree_unflatten, _broadcast_to_and_flatten, TreeSpec
+from torch.utils._pytree import tree_flatten, tree_unflatten, _broadcast_to_and_flatten, TreeSpec, _register_pytree_node
 from .pytree_hacks import tree_map_
 from functools import partial
 import inspect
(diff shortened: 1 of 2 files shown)


functorch/_src/vmap.py
```diff
@@ -40,8 +41,20 @@ def register_torch_return_types():
...
+# Temporary OrderedDict registration as pytree
+def _odict_flatten(d):
+    return list(d.values()), list(d.keys())
+
+
+def _odict_unflatten(values, context):
+    return OrderedDict((key, value) for key, value in zip(context, values))
+
+
+_register_pytree_node(OrderedDict, _odict_flatten, _odict_unflatten)
```
zou3519, line 54: Can you add a really quick test somewhere so we don't accidentally regress on this? Something that tests (1) calling flatten returns the values, (2) calling flatten than unflatten returns an equivalent odict should be sufficient.

zou3519 (approved):
from code reading this looks correct to me. It would be nice to have a small test case to avoid regressing on this accidentally. In the long run we want to upstream this into pytorch/pytorch where it will have more aggressive testing

zou3519 (approved):

vfdev-5:
@zou3519 can we merge this PR, CI is sucessful but not updated here: 
> GitHub rate limited us while adding a commit status of success: API rate limit exceeded for user ID 6422482

zou3519:
merged!

PR #563, opened 2022-03-04
Docs website modification

zou3519:
I modified the docs website:
- It has a new install page which we'll fill out l ater
- It has a new "whirlwind tour" section which is just our README
copy-pasted
- It has a "UX Limitations" section describing arbitrary python
mutation, in-place operations, and control flow.

Please let me know if you think anything else should be here.

Future work:
- I might put up a table of what operators we have support for and
which ones we don't, autogenerated via OpInfo

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -11,9 +11,13 @@ It aims to provide composable vmap and grad transforms that work with PyTorch mo
 and PyTorch autograd with good eager-mode performance.
 
 .. note::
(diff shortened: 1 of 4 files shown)


zou3519:
For the docs preview: when the `Build Docs / build-docs (pull_request)` job is done building, you can click on it, go to "Summary", and download "docs.whl". Unzipping docs.whl gives the html files that can then be opened in a web browser.

PR #566, opened 2022-03-07
Add randint_like support (should be final randomness function! 🥳)

samdow:


test/test_vmap.py
```diff
@@ -3590,7 +3575,7 @@ def test_feature_dropout(self, device, randomness):
             vmap_result = vmap(op, randomness=randomness)(passed)
             torch.manual_seed(seed)
             if randomness == 'different':
-                expected = op(passed.flatten(0, 1))
+                expected = op(passed.flatten(0, 1)) # (B0, B, ...) -> (B0 * B, ...)
```
zou3519, line 3578: flake8 alert (not sure if it's this line, just needed to put the comment somewhere)

functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -149,6 +149,33 @@ Tensor unary_pointwise_random_batch_rule(const Tensor& tensor, ExtraArgs... extr
   return makeBatched(out, 0, cur_level);
 }
 
+template<typename... ExtraArgs>
+Tensor randint_like_random_batch_rule(const Tensor& self, ExtraArgs... extra_args) {
```
zou3519, line 153: @samdow -- how do we handle randn_like, rand_like right now?

functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -149,6 +149,33 @@ Tensor unary_pointwise_random_batch_rule(const Tensor& tensor, ExtraArgs... extr
...
+  RandomnessType randomness = maybe_layer->randomness();
+  check_randomness(randomness);
+
+  Tensor tensor_value;
+  optional<int64_t> tensor_bdim;
+  std::tie(tensor_value, tensor_bdim) = unwrapTensorAtLevel(self, cur_level);
+  tensor_value = moveBatchDimToFront(tensor_value, tensor_bdim);
+
+  if (randomness == RandomnessType::Same && tensor_bdim) {
+    tensor_value = tensor_value[0];
```
zou3519, line 167: edge case, but what if tensor_vlaue has no dimensions? `tensor_value = torch.ones({})`

functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -149,6 +149,33 @@ Tensor unary_pointwise_random_batch_rule(const Tensor& tensor, ExtraArgs... extr
...
+  } else if (randomness == RandomnessType::Different && !tensor_bdim) {
+    auto shape = tensor_value.sizes();
+    VmapDimVector shapeVec(shape.begin(), shape.end());
+    shapeVec.insert(shapeVec.begin(), maybe_layer->batchSize());
```
zou3519, line 171: Nit: this insertion is O(n), we should insert first, then copy over the rest of the list

zou3519: Also, shapeVec isn't being used?

test/test_vmap.py
```diff
@@ -3581,7 +3566,7 @@ def test_feature_dropout(self, device, randomness):
         ]
 
         for op in supported_ops:
-            passed = torch.ones([4, 3, 3, 3, 3])
+            passed = torch.ones([4, 3, 3, 3, 3], device=device)
```
zou3519, line 3569: We should really test batch_dim != 0 as well as batch_dim == 0 (IIRC all the existing tests test batch_dim == 0)?

zou3519: While I'm on the "things to improve testing train": we should have a one-off manual test with just a single op that tests multiple vmaps and different randomness behavior to check that the mode mechanism works with multiple levels.

That is, we should check `vmap(vmap(f, randomness=different), randomness=same)` in all 9 combinations of randomness = {same, different, error} x {same, different, error}.

functorch/csrc/VmapModeRegistrations.cpp
```diff
@@ -5,39 +5,15 @@
...
 
 
 namespace at {
 namespace functorch {
 
-void unsupportedRandomOp(const c10::OperatorHandle& op, torch::jit::Stack* stack) {
-  TORCH_CHECK(false, "vmap: We do not yet support calling random operations inside of vmap. ",
-              "Please perform random operations outside of vmap as a workaround");
-}
-
```
zou3519, line 23: (no action required) I can see some future where when someone adds a new randomness function to native_functions.yaml, we automatically generate an "unsupported" error message. But we can bring this back when that happens

zou3519 (commented):
some questions & testing suggestions

functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -149,6 +149,33 @@ Tensor unary_pointwise_random_batch_rule(const Tensor& tensor, ExtraArgs... extr
...
+    tensor_value = tensor_value.unsqueeze(0).expand(batch_size);
```
zou3519, line 172: Is this assuming `tensor_value` has 0 dimensions?

functorch/csrc/BatchRulesRandomness.cpp
samdow: Good point--didn't see those in the original randomness error message but I'll double check them and probably end up adding them here

functorch/csrc/BatchRulesRandomness.cpp
samdow: Would we be able to vmap over a tensor with no dimensions?

functorch/csrc/BatchRulesRandomness.cpp
samdow: Good call--I think this was supposed to use shapeVec and now I'm a little confused how it works (probably another sign to reassess the randomness testing)

samdow:
Going to put in a fix for dropout running on an unbatched input. This is going to stack on top of that so this will be paused (but will fix shapeVec usage + add other like functions while the dropout fix is testing + landing)

functorch/csrc/BatchRulesRandomness.cpp
zou3519: Aha, good point. You're right -- this won't happen and tensor_value will have at least one dimension

PR #574, opened 2022-03-08
made some modifications to work with new OpOverloadPacket changes

Chillee:


Chillee:
Uh whoops, i didn't actually include my changes.

PR #575, opened 2022-03-08
Fix + test dropout without batch dimension

samdow:
After trying to update some of the tests in `randint_like` after Richard pointed out that it was wrong for the case where an incoming tensor is unbatched, found out that dropout wasn't dealing with that case correctly. If we vmap over dropout that takes in an regular tensor using different randomness, we'll need to upgrade it to a batched tensor that's expanded to have the correct behavior

### DIFF
--- a/functorch/csrc/BatchRulesRandomness.cpp
+++ b/functorch/csrc/BatchRulesRandomness.cpp
@@ -165,9 +165,6 @@ std::tuple<Tensor,Tensor> native_dropout_batching_rule(const Tensor& tensor, dou
   }
 
   if ((train.has_value() && !train) || randomness == RandomnessType::Different) {
-    if (!tensor_bdim) {
-      tensor_value = tensor_value.unsqueeze(0).expand(maybe_layer->batchSize()); // ret value will be a batched tensor
-    }
     auto res = at::native_dropout(tensor_value, p, train);
     return std::make_tuple(makeBatched(std::get<0>(res), 0, cur_level), makeBatched(std::get<1>(res), 0, cur_level));
   }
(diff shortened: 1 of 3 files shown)


functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -399,6 +425,11 @@ TORCH_LIBRARY_IMPL(aten, FuncTorchVmapMode, m) {
   UNARY_POINTWISE_RANDOM(poisson);
   UNARY_POINTWISE_RANDOM(bernoulli);
 
+  DROPOUT_UNSQUEEZE(dropout);
+  DROPOUT_UNSQUEEZE(feature_dropout);
+  DROPOUT_UNSQUEEZE(alpha_dropout);
+  DROPOUT_UNSQUEEZE(feature_alpha_dropout);
```
zou3519, line 431: These are composite operations, right? Registering a batching rule for composite operations is generally not a good idea.

Consider the following:
```
def f(x):
  y = F.dropout2d(x)
  return y.sum()

vmap(grad(f))(x)
```
This will construct a GradTensor(BatchedTensor(x)). Calling F.dropout2d( GradTensor(BatchedTensor(x))):
- will eventually make it to at::dropout(GradTensor(BatchedTensor(x)))
- GradTensor will decompose at::dropout into its constituents and BatchedTensor will not get a chance to apply the batching rule you've defined for it.

I still don't really understand what dropout is doing, so let me go read up on that a little more

zou3519 (commented):
Registering a rule for a composite operation is generally not a good idea because the composite operation can get decomposed before hitting the vmap subsystem. Maybe it's OK in this case though, it really depends on the behavior of dropout (I need to go think about this some more)

functorch/csrc/BatchRulesRandomness.cpp
zou3519, line 431: Okay, I thought about it some more. Let's talk about the case where input isn't batched but the user wanted different randomness.

So assume input has shape (3,) and we are vmapping over a dimension of size B.

For `at::dropout(input)`:
- at::native_dropout(input) route: This is fine. We expand the input to (B, 3) and then pass it through at::native_dropout.
- for `_dropout` route: we've got a problem, https://github.com/pytorch/pytorch/blob/97b20b9b50b5b1460c7284438184901c6eecaa3c/aten/src/ATen/native/Dropout.cpp#L58-L59. at::empty_like(input) produces a Tensor of size (3,), but then we call `bernoulli_` on it :/. This raises an error because we cannot do in-place on this randomness pattern.

Am I analyzing this correctly? If so, this implies we'll need to change pytorch/pytorch's implementation of at::dropout, perhaps by introducing an out-of-place at::bernoulli operation.

functorch/csrc/BatchRulesRandomness.cpp
```diff
@@ -182,6 +179,32 @@ std::tuple<Tensor,Tensor> native_dropout_batching_rule(const Tensor& tensor, dou
...
+  Tensor input_value;
+  optional<int64_t> input_bdim;
+  std::tie(input_value, input_bdim) = unwrapTensorAtLevel(input, cur_level);
+
+  if (!input_bdim && randomness == RandomnessType::Different) {
+    VmapDimVector shapeVec(1, maybe_layer->batchSize());
+    auto shape = input.sizes();
+    shapeVec.reserve(input_value.dim() + 1);
+    shapeVec.insert(shapeVec.end(), shape.begin(), shape.end());
+    input_value = input_value.unsqueeze(0).expand(shapeVec);
```
zou3519, line 201: nit: do we actually need the unsqueeze(0)?

PR #585, opened 2022-03-09
Compile readme

anijain2305:
Added some description at the top and added doc and tutorial links.

I have not gone through major overhaul here. Not sure if its required.

@Chillee @zou3519


zou3519 (approved):
lgtm

PR #587, opened 2022-03-10
Add website note on randomness

samdow:
Adds the following to the bottom of the "UX Limitations" page of the website

Screen Shot 2022-03-11 at 3 15 59 PM
Screen Shot 2022-03-11 at 3 16 11 PM

### DIFF
--- a/docs/source/ux_limitations.rst
+++ b/docs/source/ux_limitations.rst
@@ -179,3 +179,42 @@ JAX supports transforming over
 using special control flow operators (e.g. ``jax.lax.cond``, ``jax.lax.while_loop``).
 We're investigating adding equivalents of those to functorch
 (open an issue on `GitHub <https://github.com/pytorch/functorch>`_ to voice your support!).
+
+Randomness
+----------
+Our randomness differs from JAX because we aren't using a stateless PRNG, in part because PyTorch doesn't
+have full support for a stateless PRNG. Instead, we've introduced a flag system to allow for the most
+common forms of randmoness that we see. If your use case does not fit these forms of randomness, please
+file an issue.
+
+The flag can only be passed to vmap and can take on 3 values, "error," "different," or "same," defaulting
+to error. Under "error" mode, any call to a random function will produce an error asking the user to use
+one of the other two flags based on their use case.
+
+Under "different" randomness, elements in a batch produce different random values. For instance,
+
+::
+
+  def add_noise(x):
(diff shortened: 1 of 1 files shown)


docs/source/ux_limitations.rst
```diff
@@ -179,3 +179,45 @@ JAX supports transforming over
...
+
+Randomness
+----------
+Note that like with JAX's PRNG system, our system only determine the randomness behavior of PyTorch
+operators
+
+Our randomness differs from JAX because we aren't using a stateless PRNG, in part because PyTorch doesn't
+have full support for a stateless PRNG. Instead, we've introduced a flag system to allow for the most
+common forms of randmoness that we see. If your use case does not fit these forms of randomness, please
+file an issue.
```
zou3519, line 191: I feel like people reading this might not know what JAX's PRNG system is, so leading with it without discussing what it is might be confusing. There's a couple of main point I think we want to get across:
1. randomness inside vmap is tricky
2. we allow the user to control random PyTorch operations only (e.g. we cannot control python randomness, numpy randomness)
3. How to use the randomness API
4. Discussion of how this is different or similar to JAX's system
Would it make sense to move the discussion on JAX to the end?

docs/source/ux_limitations.rst
```diff
@@ -179,3 +179,45 @@ JAX supports transforming over
...
+
+::
+
+  def add_noise(x):
+    y = torch.randn(4)  # y will be different across the batch
+    return x + y
+
+  x = torch.ones(3, 4)
+  result = vmap(add_noise, randomness="different")(x)
+  assert result[0] != result[1]
```
zou3519, line 207: would this be easier to understand if we did fewer operations?
```
def f(x):
  return torch.rand([])

result = vmap(f, randomness='different')(torch.ones(3))
# we get three different random numbers!
```
The same randomness case is, we get 3 same random numbers

zou3519 (commented):
I had a question on organization

docs/source/ux_limitations.rst
samdow: Makes sense! I added a little intro and moved this to the end.

docs/source/ux_limitations.rst
samdow: Agreed. Changed

zou3519 (approved):
thank you, this reads better now!

PR #593, opened 2022-03-11
Refactor randomness tests

samdow:
Refactors the randomness tests based on their input type + OOP vs in-place. So this makes a new test for:
- OOP, takes in shape
- OOP, takes in one tensor
- OOP, takes in two tensors (also tests if either tensor is batched as well as both)
- in place, takes in one tensor
- in place, takes in two tensors (this is only bernoulli, also tests if either tensor is batched as well as both)

It also leaves the randperm + dropout tests because those are special cased

### DIFF
--- a/test/test_vmap.py
+++ b/test/test_vmap.py
@@ -3466,55 +3466,74 @@ class TestVmapOperatorsOpInfo(TestCase):
         y = torch.randn(2, 3, device=device)
         self.assertTrue(isinstance(vmap(f)(x, y), Point))
 
-    def reset_random(self, generator, orig_state, use_generator, seed):
+
+class TestRandomness(TestCase):
+    def _reset_random(self, generator, orig_state, use_generator, seed):
         return generator.set_state(orig_state) if use_generator else torch.manual_seed(seed)
+
+    def _get_image(self, batched_input, batch_size, device):
+        if batched_input:
+            return torch.ones([batch_size, 3, 3, 14, 14], device=device)
+        return torch.ones([3, 3, 14, 14], device=device)
+
+    def _assert_all_slices_equal(self, tensor):
+        expected = tensor[0]
+        self.assertTrue((tensor == expected).all())
+
+    def _assert_all_slices_unique(self, tensor):
+        B0 = tensor.shape[0]
+        slices_equal = vmap(vmap(lambda x, y: (x == y).all(), (0, None)), (None, 0))(tensor, tensor)
+        assert slices_equal.shape == (B0, B0)
+        slices_equal.diagonal().zero_()
+        self.assertEqual(slices_equal, torch.zeros_like(slices_equal))
+
+    def _assert_throws_in_error_mode(self, fn, args, in_dims=0):
+        with self.assertRaisesRegex(RuntimeError, r"called random operation while in randomness error mode"):
+            vmap(fn, in_dims=in_dims, randomness="error")(*args)
+
+    def _assert_throws_in_different_mode_inplace(self, fn, args, in_dims=(None, 0)):
+        with self.assertRaisesRegex(RuntimeError, r"different inplace randomness on an unbatched tensor"):
+            vmap(fn, in_dims=in_dims, randomness="different")(*args)
+
+    def _in_dims(self, *batched):
+        batched = batched + (True,)  # for the always batched dummy argument
+        return tuple(0 if is_batched else None for is_batched in batched)
 
     @parametrize('randomness', ['same', 'different', 'error'])
     @parametrize('use_generator', [True, False])
-    def test_random_behavior(self, device, randomness, use_generator):
+    def test_shape_OOP_ops(self, device, randomness, use_generator):
+        # ops that take in the final shape and are out of place
         generator = torch.Generator(device=device)
         orig_state = generator.get_state()
         kwargs = {'device': device, 'generator': generator} if use_generator else {'device': device}
-        only_gen_kwarg = {'generator': generator} if use_generator else {}
-        supported_random_ops = [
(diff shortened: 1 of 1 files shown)


test/test_vmap.py
```diff
@@ -3466,57 +3466,81 @@ def f(x, y):
...
+            vmap(fn, in_dims=in_dims, randomness="different")(*args)
+
+    def _assert_throws_in_same_mode_batched(self, fn, args, in_dims=0):
+        with self.assertRaisesRegex(RuntimeError,
+                                    r"Vmap does not currently support same randomness with a batched tensor input"):
+            vmap(fn, in_dims=in_dims, randomness="same")(*args)
+
+    def _in_dims(self, *batched):
+        batched = batched + (True,)  # for the always batched dummy argument
+        return tuple(0 if is_batched else None for is_batched in batched)
```
zou3519, line 3505: Doesn't need to happen in this PR, but we should test a case where, if the tensor already has a dimension being vmapped over, that dimension is not at the front of the tensor (so, not zero)

test/test_vmap.py
```diff
@@ -3466,57 +3466,81 @@ def f(x, y):
...
+
     @parametrize('randomness', ['same', 'different', 'error'])
     @parametrize('use_generator', [True, False])
-    def test_random_behavior(self, device, randomness, use_generator):
+    def test_shape_OOP_ops(self, device, randomness, use_generator):
+        # ops that take in the final shape and are out of place
```
zou3519, line 3510: nit: oop is generally the abbreviation for "object-oriented-programming". By "shape oop" you probably mean "factory" -- a factory function is a function that accepts no Tensor arguments and returns a Tensor. Maybe let's call it `test_factory_ops` ?

test/test_vmap.py
```diff
@@ -3746,49 +3754,206 @@ def test_like_functions(self, device, randomness, batched_input):
 
     @parametrize('use_generator', [True, False])
     @parametrize('randomness', ['error', 'same', 'different'])
-    def test_random_inplace_not_batched(self, device, use_generator, randomness):
-        # tests that when in place random ops are being called during vmap but not with a batched tensor,
-        # it behaves like a normal random_ when using same random and errors during different random
-        supported_ops = [
+    @parametrize('batched_input', [True, False])
+    def test_random_inplace_single(self, device, use_generator, randomness, batched_input):
```
zou3519, line 3758: nit: test_random_unary_inplace: "unary" is shorthand for "single tensor input"

test/test_vmap.py
```diff
@@ -3746,49 +3754,206 @@ def test_like_functions(self, device, randomness, batched_input):
...
+        orig_state = generator.get_state()
+        kwargs = {'generator': generator} if use_generator else {}
+        p = torch.rand(14)
+        ops = [
             lambda t, _: t.random_(**kwargs),
             lambda t, _: t.random_(100, **kwargs),
             lambda t, _: t.random_(-5, 100, **kwargs),
             lambda t, _: t.normal_(**kwargs),
-            lambda t, _: t.bernoulli_(torch.tensor([0.3, 0.4, 0.5, 0.6]), **kwargs),
+            lambda t, _: t.bernoulli_(p, **kwargs),
```
zou3519, line 3768: this is not really a single Tensor input, right? Because t is a Tensor and p is also a Tensor. Should this be here?

test/test_vmap.py
```diff
@@ -3746,49 +3754,206 @@ def test_like_functions(self, device, randomness, batched_input):
...
+            expected = op(input_expected, probability, always_batched)
+            self._assert_all_slices_equal(vmap_result)
+            for i in range(B0):
+                self.assertEqual(vmap_result[i], expected)
+
+    @parametrize('use_generator', [True, False])
+    @parametrize('randomness', ['error', 'same', 'different'])
+    @parametrize('batched_input', [True, False])
+    @parametrize('batched_other', [True, False])
+    def test_random_out_of_place_two_tensors(self, device, use_generator, randomness, batched_input, batched_other):
```
zou3519, line 3861: nit: test_binary_outplace -- "binary" is the official term for a function that accepts two tensor arguments. But also feel free to ignore, it's pretty clear from this test name what is going on

zou3519 (approved):
The tests look great after the cleanup!

One minor nit: on vocabulary: "unary" can be shorthand for "function that accepts single Tensor argument", "binary" can be shorthand for "function that accepts two Tensor arguments", "factory" can be shorthand for "function that accepts no Tensor arguments but returns a Tensor". Your tests names are already pretty descriptive so feel free to ignore.

Next, it would be nice if we tested a tensor where the dimension being vmapped over isn't the front (right now it's hardcoded in as (B, 3, 3, 14, 14). But this doesn't need to happen in this PR (it's better as a follow-up)

test/test_vmap.py
samdow: Good catch! Not certain how that happened but removed this and added kwargs to the bernoulli_in_place test that does take in a tensor for probability

test/test_vmap.py
samdow: Fixed

test/test_vmap.py
samdow: Fixed

test/test_vmap.py
samdow: Fixed

test/test_vmap.py
samdow: Sounds good! I'll get another PR for that. There will be another follow up since this already raised a bug with multinomial so good to check these things

zou3519 (approved):

PR #594, opened 2022-03-11
Make a number of small changes to randomness

samdow:
These were from @zou3519's comments on the PRs that I missed as I was trying to get everything in. It's ~4 tiny changes split up into commits:
- fixes 2 instances of `shapeVec` that were still doing O(n) insertion
- removes unnecessary unsqueeze
- adds comment explaining the cases in binary randomness
- corrects the inplace error message to say bernoulli, not arithmetic


zou3519 (approved):
LGTM, thanks for improving the code.

~~I forget, do the randomness tests test inputs where the dimension being vmapped over is not at the very front of the Tensor? That's important to test if we haven't done that yet -- I caught a lot of bugs in the regular batching rules after adding a variant of that to test_vmap_exhaustive~~ let me make this comment after reading through your test refactor PR

PR #609, opened 2022-03-21
Prelu batching rule (forward + backward)

samdow:
Adds the batching rule for prelu. Will need to wait on merging of [prelu opinfo fix](https://github.com/pytorch/pytorch/pull/74467) in order to get rid of the last commit that does a hacky fix so that the weight is passed as a positional arg and is therefore batchable

### DIFF
--- a/functorch/csrc/BatchRulesActivation.cpp
+++ b/functorch/csrc/BatchRulesActivation.cpp
@@ -8,7 +8,8 @@
 #include <functorch/csrc/PlumbingHelper.h>
 #include <ATen/Operators.h>
 
-// NB: most activation functions are pointwise unary or binary. These are the ones that have special batch rules
+// NB: most activation functions fit pointwise unary or binary rules.
+// These are only the ones that have special batch rules to help with organization
 namespace at { namespace functorch {
 std::tuple<Tensor,optional<int64_t>> prelu_batch_rule(
     const Tensor& input, optional<int64_t> input_bdim,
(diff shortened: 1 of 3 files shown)


functorch/csrc/BatchRulesBinaryOps.cpp
```diff
@@ -275,6 +275,140 @@ std::tuple<Tensor,optional<int64_t>> cdist_backward_batch_rule(
...
+  VmapDimVector new_shape(1, batch_size);
+  new_shape.reserve(input.dim() + 1);
+  new_shape.insert(new_shape.end(), input.sizes().begin(), input.sizes().end());
+  return new_shape;
+}
+
+std::tuple<Tensor,optional<int64_t>,Tensor,optional<int64_t>> prelu_backward_batch_rule(
+    const Tensor& grad_out, optional<int64_t> grad_out_bdim,
+    const Tensor& self, optional<int64_t> self_bdim,
+    const Tensor& weight, optional<int64_t> weight_bdim) {
```
samdow, line 346: move this out of binary since it's not

functorch/csrc/BatchRulesActivation.cpp
```diff
@@ -0,0 +1,180 @@
...
+  } else { // !need_bdim && !has_bdim
+    VmapDimVector new_shape(input.sizes().begin(), input.sizes().end());
+    return new_shape;
+  }
+}
+
+std::tuple<Tensor, Tensor> prelu_backward_batched(
+    const Tensor& grad_out, const Tensor& self, const Tensor& weight,
+    const VmapDimVector& self_shape_with_bdim, const VmapDimVector& padded_shape, const VmapDimVector& unpadded_shape) {
+  // helper function that assumes that self, weight, and grad_out are all of the correct shape
```
samdow, line 100: incorrect comment since we're no longer ensuring the shape

functorch/csrc/BatchRulesActivation.cpp
```diff
@@ -0,0 +1,180 @@
...
+    return new_shape;
+  } else { // !need_bdim && !has_bdim
+    VmapDimVector new_shape(input.sizes().begin(), input.sizes().end());
+    return new_shape;
+  }
+}
+
+std::tuple<Tensor, Tensor> prelu_backward_batched(
+    const Tensor& grad_out, const Tensor& self, const Tensor& weight,
+    const VmapDimVector& self_shape_with_bdim, const VmapDimVector& padded_shape, const VmapDimVector& unpadded_shape) {
```
samdow, line 99: self_grad_shape, weight_grad_padded_shape, weight_grad_shape

PR #612, opened 2022-03-22
[WIP] [Do not merge] Reduce overhead of AOT Module

anijain2305:
@jansel @Chillee 

Inspired heavily from - https://github.com/facebookresearch/torchdynamo/blob/ce0c84a62d3287e2afde22e0d823a8d1ae4758a8/torchdynamo/optimizations/python_key.py#L121

### DIFF
--- a/functorch/_src/aot_autograd.py
+++ b/functorch/_src/aot_autograd.py
@@ -487,52 +487,7 @@ def clear_compile_cache():
         compile_cache = None
 
 
-# def aot_module(mod: nn.Module, *args, **kwargs) -> nn.Module:
-#     """
-#     Traces the forward and backward graph of :attr:`mod` using torch dispatch
-#     tracing mechanism. It is wrapper function, that underneath uses
-#     :func:`aot_function` to perform tracing and compilation.
-# 
-#     :func:`aot_module` lifts the parameters and buffers of ``nn.Module`` as inputs
-#     to a new callable which is then compiled through :func:`aot_function`.
-# 
-#     .. warning::
-#         This API is experimental and likely to change.
-# 
-#     Args:
-#         mod (Callable): A ``nn.Module`` module.
-#         args : args to be passed to :func:`aot_function`
-#         kwargs : kwargs to be passed to :func:`aot_function`
-# 
-#     Returns:
-#         Returns a ``nn.Module`` that retains the eager behavior of the original
-#         :attr:`mod`, but with forward and backward graph compiled.
-# 
-#     """
-# 
-#     def functional_call(named_params, named_buffers, *args, **kwargs):
-#         params_and_buffers = {**named_params, **named_buffers}
-#         return _stateless.functional_call(mod, params_and_buffers, args, kwargs)
-# 
-#     compiled_f = aot_function(functional_call, *args, **kwargs)
-# 
-#     class AOTModule(nn.Module):
-#         def __init__(self):
-#             super(AOTModule, self).__init__()
-#             self.orig_module = mod
-# 
-#         def forward(self, *args, **kwargs):
-#             return compiled_f(
-#                 dict(_named_parameters(mod, remove_duplicate=False)),
-#                 dict(_named_buffers(mod, remove_duplicate=False)),
-#                 *args,
-#                 **kwargs,
-#             )
-# 
-#     return AOTModule()
-
-
-def aot_module(mod: nn.Module, *top_args, **top_kwargs) -> nn.Module:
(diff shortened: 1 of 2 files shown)


functorch/_src/aot_autograd.py
```diff
@@ -506,24 +555,119 @@ def aot_module(mod: nn.Module, *args, **kwargs) -> nn.Module:
...
+    #########################################################
+    """
+    (1) Create a new fn_for_tracing that lifts params as inputs (TODO: buffers)
+    (2) A new tracer - MyTracer (slight modification of PythonKeyTracer). This
+    works with Proxy tensors instead of PythonTensors. Goal is to get a torch
+    graph, and not torch.aten graph yet.
+    (3) This traced function is then passed on to aot_function.
+    (4) The params are read and flattened on every forward call, as they can change during training.
+    """
+    #########################################################
```
anijain2305, line 568: @jansel @Chillee This is the flow.

functorch/_src/aot_autograd.py
```diff
@@ -12,6 +12,8 @@
 from .named_members_polyfill import _named_parameters, _named_buffers
 from typing import Callable, List, Dict, Any, Tuple, Optional
 
+import torchdynamo
```
jansel, line 15: perhaps:
```
try:
    from torchdynamo import disable as disable_dynamo
except ImportError:
    def disable_dynamo(x):
         return x
```

So we can still use this without dynamo.

functorch/_src/aot_autograd.py
```diff
@@ -506,24 +555,119 @@ def aot_module(mod: nn.Module, *args, **kwargs) -> nn.Module:
...
+        traced = torch.fx.GraphModule(tracer.root, graph, "my_tracer")
+        traced.recompile()
+
+
+        def fn_with_params_as_args(*joint_args):
+            return traced(*joint_args)
+        return fn_with_params_as_args
+
+
+    gm = torch.fx.symbolic_trace(mod)
```
jansel, line 641: Why do we need this?  Seems not ideal, as this could be lossy.

functorch/_src/aot_autograd.py
```diff
@@ -506,24 +555,119 @@ def aot_module(mod: nn.Module, *args, **kwargs) -> nn.Module:
...
+            params = {
+                **dict(_named_parameters(gm, remove_duplicate=False)),
+                **dict(_named_buffers(gm, remove_duplicate=False)),
+            }
+            if HAS_TREE:
+                params_flat = tree.flatten(params)
+            else:
+                params_flat, _ = pytree.tree_flatten(params)
+
+            params_flat = tuple(params_flat)
```
jansel, line 662: In the torchdynamo case (not in the general case), this list is a constant you don't need to recompute every time.

functorch/_src/aot_autograd.py
```diff
@@ -506,24 +555,119 @@ def aot_module(mod: nn.Module, *args, **kwargs) -> nn.Module:
...
+            joint_args = (*params_flat, *args) 
+            nargs = len(joint_args)
+
+            if compiled_f is None:
+                fn_with_params_as_args = flattened_fn(gm, nargs)
+                compiled_f = aot_function(fn_with_params_as_args, *top_args, **top_kwargs)
```
jansel, line 668: In the torchdynamo case (not in the general case), there should only every be one `compiled_f`.  You can just compute it ahead of time and don't need to wait until the first call.

functorch/_src/aot_autograd.py
anijain2305: For training, the params will change after each update. Don't we have to flatten for each call?

functorch/_src/aot_autograd.py
anijain2305: Ok. Yeah, we can send example_inputs and compile it beforehand.

jansel: I think there is also a bunch of pytree overhead inside aot_function, plus an unneeded caching layer.

functorch/_src/aot_autograd.py
anijain2305: Yeah, I realized while driving today. These are just references. And the update happens in place.

Will remove it.

anijain2305:
Closing. Following PRs are covering this

* Contiguous - https://github.com/pytorch/functorch/pull/536
* AOT module simplified - https://github.com/pytorch/functorch/pull/660
* Disabling torchdynamo - https://github.com/pytorch/functorch/pull/662

PR #619, opened 2022-03-23
Removing the hack to fix the avg_pool2d backward

anijain2305:
As title


anijain2305:
https://github.com/pytorch/pytorch/pull/74768 should fix this in PyTorch core itself

PR #620, opened 2022-03-25
make_functional parameter sharing fix

zou3519:
Test Plan:
- a lot of new tests

### DIFF
--- a/functorch/_src/make_functional.py
+++ b/functorch/_src/make_functional.py
@@ -56,7 +56,53 @@ def raise_parameter_tying_error():
         "https://github.com/pytorch/functorch/issues/446")
 
 
-def extract_weights(mod: nn.Module) -> Tuple[Tuple[Tensor, ...], List[str]]:
+def create_names_map(named_params, tied_named_params):
+    """
+    named_params is a dictionary of tensors: {'A': A, 'B': B}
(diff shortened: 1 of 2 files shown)


MichaelArbel:
Thanks a lot for providing the fix!
It seems that one can access a parameter of a module by:  param =  module.parameter_name which returns the value of the parameter that was provided as input to the function constructed using make_functional.

However, this approach requires knowing the parameter_name. What would be great is to be able to access the parameters using the module.parameters(). However, this returns an empty generator when using the functional form of the module. Do you think there is a workaround this?

Thanks!

zou3519:
@MichaelArbel do you have a code example? I'm wonder what `module` you're calling `.parameters()` on.

MichaelArbel:
Hi, 

I think I found a workaround by setting the weights as attributes in MyModule. It would be great to have a method that handles this automatically, but this seems to do the job so far.  Thanks again! 

```
class MyModule(nn.Module):
	def __init__(self,linear):
		super(MyModule,self).__init__()
		self.linear = linear
		weights = list(self.linear.parameters())
		
		for i,w in enumerate(weights):
			setattr(self,'weight_'+str(i),w)

	def forward(self,x):
		loss = torch.sum(self.linear(x)**2)
		regularization = torch.sum(self.weight_0**2)
		return loss + regularization

```

Issue #626, opened 2022-03-28
Issue with `tensor.numpy()` for wrapped tensors


vfdev-5:
Calling `.numpy()` on wrapped tensors, e.g. `GradTrackingTensor`, `BatchedTensor`

```
RuntimeError: Cannot access data pointer of Tensor that doesn't have storage
```

### How to reproduce

```python
import torch
import functorch as ft

def foo(t):
    tt = t.detach()
    n = tt.numpy()
    a + n.sum()
    return t

x = torch.rand(4, 3)
out = ft.grad(foo)(x)
# or
# out = ft.vmap(foo)(x)
```

PR #627, opened 2022-03-29
Fixing tensor.numpy on wrapped tensors 

vfdev-5:
Fixes https://github.com/pytorch/functorch/issues/626

Description:
- Fixing tensor.numpy on wrapped tensors
- Added a test

### DIFF
--- a/test/test_eager_transforms.py
+++ b/test/test_eager_transforms.py
@@ -863,6 +863,38 @@ class TestGradTransform(TestCase):
                 expected = expected.replace("\n", "").replace("  ", "")
                 self.assertEqual(expected, buf)
 
+    @parametrize("op_list_data", [
+        subtest(([vmap, ], [(4, 2), (64, 3, 32, 32)]), name='vmap'),
+        subtest(([vmap, vmap], [(4, 3, 2), (64, 3, 32, 32)]), name='vmap_vmap'),
+        subtest(([grad, ], [(0, ), [], (4, 2), (64, 3, 32, 32)]), name='grad'),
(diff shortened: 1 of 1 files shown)


functorch/_src/monkey_patching.py
```diff
@@ -98,3 +98,34 @@ def _backward(*args, **kwargs):
...
+    dl_enabled = _C.tls_set_is_included()
+    try:
+        # Disable temporarily kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+        if (dl_enabled):
+            _C._set_dynamic_layer_keys_included(False)
+        return value.numpy()
+    finally:
+        # Reenable kDynamicLayerFrontModeKey/kDynamicLayerBackModeKey as included dispatch keys
+        if (dl_enabled):
+            _C._set_dynamic_layer_keys_included(True)
```
zou3519, line 128: Hmm, so this is a little more complicated than this I think.

When someone calls .numpy() under vmap, we probably want to error out. Otherwise some weird things might happen:
```
def f(x):
  return torch.tensor(x.numpy())

x = torch.randn(B)
vmap(f)(x) # returns a Tensor of size B, B -- is that what we want?
```

zou3519: When someone calls .numpy() under the grad transform then we should support this (as long as there are no vmaps involved). I'm not sure what the best way to support this is... one thing we can do is keep unwrapping the Tensor and seeing that no BatchedTensors are involved.

In the long-term we want a better fix for this that perhaps involves making the pytorch dispatcher recognize .numpy() as an operation

PR #640, opened 2022-03-30
Fix gelu_backward

samdow:
Apparently gelu_backward doesn't broadcast well on mkldnn which is why we were seeing the other error


Chillee:
This feels like something that should be fixed in PyTorch 🤔

PR #644, opened 2022-03-30
switch decompositions over to use public PyTorch API instead of aten ops

Chillee:
cc: @gchanan

### DIFF
--- a/functorch/_src/decompositions.py
+++ b/functorch/_src/decompositions.py
@@ -1,4 +1,5 @@
 import torch
+import torch.nn.functional as F
 from torch import Tensor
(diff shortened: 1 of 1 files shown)


Chillee:
@anijain2305 These are currently run while tracing, so it doesn't matter what we call in the middle - just what our eventual tracing objects.

For example, if we call a decomposition for `aten.foo`, it'll look like this
```
aten.foo => torch.bar => aten.bar
```

The main advantage of this is to 1. start to align our aten operators in terms of our public API, and 2. allow us to re-use primtorch decomposition (which are written from the Python torch API).

PR #668, opened 2022-04-06
[Release] fix gelu_backward

samdow:


samdow:
Closing, this actually isn't an issue in the old version because the new gelu isn't in so 1.11 uses the infinitely differentiable version

PR #674, opened 2022-04-07
Fix the decomposition of native_layer_norm_backward operation.

gprateek93:
The return type for native_layer_norm_backward is a tuple of
optional tensors as the output_mask decides whether or not the
gradient of a particular input is to be calculated or not.

Signed-Off-By: Prateek Gupta <prateek@nod-labs.com>


Chillee:
LGTM!

Chillee (approved):

PR #676, opened 2022-04-07
Skip extracting meta tensor info for sparse tensors

anijain2305:
As title


Chillee:
I am ... a little bit worried whether this'll affect the recomputation algorithm. Probably doesn't matter.

Chillee (approved):

PR #691, opened 2022-04-13
Add Python 3.10 config to CI

zou3519:


zou3519:
Config works, it just looks like we have preexisting test failures

Issue #639, opened 2022-03-30
Memory Leak


tcapelle:
Hello!
I am thrilled with the functorch package, and have been playing with it lately.

With @soumik12345 we found a memory leak after training a NN. We documented our findings here:

http://wandb.me/functorch-intro

We are probably doing something wrong, but the memory increases after each epoch.

<img width="940" alt="image" src="https://user-images.githubusercontent.com/18441985/160872930-93dd11d0-c5e1-444f-b60d-02484f14a102.png">

As the GPU is pretty monstrous we didn't notice this straight away, but it clearly fills up progresively. The stateful pytorch training loop does not produce this.

samdow:
@tcapelle Thanks for the issue and the detailed repro! We're definitely looking at it now, but don't see anything immediately wrong. You are using the API as intended 😄 

One question for you two since it uses the `wandb` library, do you know if that samples the GPU utilization based on time or at certain calls?

samdow:
@tcapelle Thanks for the issue! Basically what's happening is that because the params in the model have `requires_grad=True`, they'll hold onto the backprop graph and keep adding to it instead of releasing that memory. Since your example doesn't need higher order gradients, we don't need to do this. There's a couple options here:

(1) make params not require_grad
```
functional_model, params = make_functional(model)
for param in params:
    param.requires_grad_(False)
```

(2) use detach in sgd
```
# naive sgd optimizer implementation
def sgd_optimizer(weights, gradients, learning_rate):
    def step(weight, gradient):
        return weight - learning_rate * gradient

    assert (len(weights) == len(gradients))
    return [step(weight, gradient.detach()) for weight, gradient in zip(weights, gradients)]
```

tcapelle:
I was looking for something like this, zero_grad or detach, but didn't know where to do it.

Then a training tutorial should showcase this. I can work out with @soumik12345 an example to add.

vfdev-5:
@samdow I was wondering if solution (1) could be an optional behaviour for `make_functional(model)`, e.g. `make_functional(model, params_require_grad=False)` ? Such that users wont experience mem issue ...

tcapelle:
Ohh, I would also prefer this as default.

samdow:
> @samdow I was wondering if solution (1) could be an optional behaviour for make_functional(model), e.g. make_functional(model, params_require_grad=False) ? Such that users wont experience mem issue ...

[...]

PR #701, opened 2022-04-14
Added params_require_grad arg to make_functional*

vfdev-5:
Related to https://github.com/pytorch/functorch/issues/639#issuecomment-1097154660

Description:
- Added `params_require_grad` argument to `make_functional` and `make_functional_with_buffers`
- Added tests

@samdow let me know if something else I'm missing for this PR. Thanks

### DIFF
--- a/functorch/_src/make_functional.py
+++ b/functorch/_src/make_functional.py
@@ -257,11 +257,14 @@ class FunctionalModuleWithBuffers(nn.Module):
         self.all_names_map.update(buffer_names_map)
 
     @staticmethod
-    def _create_from(model):
+    def _create_from(model, params_require_grad=True):
         # TODO: We don't need to copy the model to create a stateless copy
         model_copy = copy.deepcopy(model)
         params, param_names, param_names_map = extract_weights(model_copy)
         buffers, buffer_names, buffer_names_map = extract_buffers(model_copy)
+        if not params_require_grad:
+            for param in params:
+                param.requires_grad_(False)
         return (
             FunctionalModuleWithBuffers(model_copy, param_names, buffer_names,
                                         param_names_map, buffer_names_map),
@@ -294,10 +297,13 @@ class FunctionalModule(nn.Module):
         self.names_map = names_map
 
     @staticmethod
-    def _create_from(model):
+    def _create_from(model, params_require_grad=True):
         # TODO: We don't need to copy the model to create a stateless copy
         model_copy = copy.deepcopy(model)
         params, param_names, names_map = extract_weights(model_copy)
+        if not params_require_grad:
+            for param in params:
+                param.requires_grad_(False)
         return FunctionalModule(model_copy, param_names, names_map), params
 
     def forward(self, params, *args, **kwargs):
@@ -310,7 +316,7 @@ class FunctionalModule(nn.Module):
             _swap_state(self.stateless_model, self.names_map, old_state)
 
 
-def make_functional(model: nn.Module):
+def make_functional(model: nn.Module, params_require_grad: bool = True):
     """make_functional(model) -> func, params
 
     Given a ``torch.nn.Module``, :func:`make_functional` extracts the state
@@ -353,15 +359,21 @@ def make_functional(model: nn.Module):
 
     If the model has any buffers, please use :func:`make_functional_with_buffers` instead.
 
+    Args:
+        model (torch.nn.Module): Input model.
+        params_require_grad (bool): If True, output parameters will have require gradient.
+            It could be helpful to set this flag to False if computions do not need higher order gradients.
+            Default: True.
+
     """
     buffers = list(model.buffers())
     if len(buffers) > 0:
         raise RuntimeError('make_functional(model): `model` has buffers. Please use '
                            'make_functional_with_buffers(model) instead.')
-    return FunctionalModule._create_from(model)
+    return FunctionalModule._create_from(model, params_require_grad=params_require_grad)
 
 
-def make_functional_with_buffers(model: nn.Module):
+def make_functional_with_buffers(model: nn.Module, params_require_grad: bool = True):
     """make_functional_with_buffers(model) -> func, params, buffers
 
     Given a ``torch.nn.Module``, make_functional_with_buffers extracts the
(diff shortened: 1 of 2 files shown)


functorch/_src/make_functional.py
```diff
@@ -310,7 +316,7 @@ def forward(self, params, *args, **kwargs):
             _swap_state(self.stateless_model, self.names_map, old_state)
 
 
-def make_functional(model: nn.Module):
+def make_functional(model: nn.Module, params_require_grad: bool = True):
```
zou3519, line 319: I have one main question about how this API works:
1. params_require_grad: should we make all parameters require grad, or should we mirror the state on the nn.Module (can a parameter on nn.Module have requires_grad=False) ?

vfdev-5: Yes, I agree, maybe this API can be confusing at this point. Probably, naming should reflect the fact that we could disable enabled grads but not enable disabled... Let me think about the naming.

samdow: Agree that we can fix this with naming, especially since @vfdev-5 (correctly) only sets the params to `requires_grad(False)` and never the other way around. Maybe (or at least as a jumping off point),`disable_params_grad` (flipping the boolean's default)?

functorch/_src/make_functional.py
```diff
@@ -353,15 +359,21 @@ def compute_loss(params, x, t):
 
     If the model has any buffers, please use :func:`make_functional_with_buffers` instead.
 
+    Args:
+        model (torch.nn.Module): Input model.
+        params_require_grad (bool): If True, output parameters will have require gradient.
+            It could be helpful to set this flag to False if computions do not need higher order gradients.
```
samdow, line 365: nit: maybe "computations do not need higher order gradients" -> "gradient computations only use the functorch API"

This point definitely got jumbled in the discussion on the issue, but basically we can still get higher order gradients with this, we just would have to use functorch's `grad(grad(...))`. The issue will be if we did something like `grad(...).backward()`

vfdev-5: @samdow could you please check again the docstring and maybe help to improve it.

functorch/_src/make_functional.py
vfdev-5: Let's make it as `disable_params_grad`

functorch/_src/make_functional.py
```diff
@@ -353,15 +359,22 @@ def compute_loss(params, x, t):
 
     If the model has any buffers, please use :func:`make_functional_with_buffers` instead.
 
+    Args:
+        model (torch.nn.Module): Input model.
+        disable_params_grad (bool): Flag to disable gradients for output parameters.
+            It could be helpful to set this flag to True if gradient computations only use the functorch API.
+            Disabling gradients also helps to avoid accumulating them in the backprop graph and to prevent
+            keeping them in memory. Default: False.
```
zou3519, line 367: This is a bit ambiguous. I'd phrase it something like the following:
- The returned params are unrelated to the set of params from the original model. If True (default), the params will have requires_grad=True on them (aka they will be trackable with regular PyTorch autograd), matching the requires_grad-ness of the params from the original model. Otherwise, the returned params will have requires_grad=False.
- If you plan on using regular PyTorch autograd (e.g., if you want to call .backward() or torch.autograd.grad(), then set disable_params_grad=True (default)). Otherwise, if you're only planning on using functorch's gradient transforms, then please set disable_params_grad=False to avoid unnecessarily tracking history with PyTorch autograd.

functorch/_src/make_functional.py
```diff
@@ -353,15 +359,22 @@ def compute_loss(params, x, t):
 
     If the model has any buffers, please use :func:`make_functional_with_buffers` instead.
 
+    Args:
+        model (torch.nn.Module): Input model.
+        disable_params_grad (bool): Flag to disable gradients for output parameters.
```
zou3519, line 364: Can we bikeshed this name a little more? The reason why it is confusing for me is that we're not really disabling gradients, what we are doing is turning on or off regular PyTorch Autograd tracking of the input Tensors.

Here are some potential names:
- set_requires_grad (requires_grad is the property that PyTorch autograd uses to track gradients)
- requires_grad
- enable_pytorch_autograd

Thoughts? @vfdev-5 @samdow

samdow: Per the other discussion, I think `set_requires_grad ` and `requires_grad ` might be confusing since parameters don't _have_ to `requires_grad` and we aren't changing them if it's True

Maybe `enable_pytorch_autograd`? @vfdev-5 what do you think?

zou3519: > disable_autograd_tracking

I like this better. Eventually we'll upstream functorch into pytorch/pytorch and "enable_pytorch_autograd" will get even more confusing: "What do you mean by PyTorch autograd?"

samdow: Same! Thanks for the suggestion @vfdev-5!

vfdev-5:
@zou3519 @samdow I updated the name and the docstring (and skipped the CI) to wait for your approval. If everything is fine, then I can restart the CI, we'll be able to merge.

functorch/_src/make_functional.py
```diff
@@ -310,7 +316,7 @@ def forward(self, params, *args, **kwargs):
             _swap_state(self.stateless_model, self.names_map, old_state)
 
 
-def make_functional(model: nn.Module):
+def make_functional(model: nn.Module, disable_autograd_tracking: bool = False):
     """make_functional(model) -> func, params
```
samdow, line 320: nit: (sorry I missed this earlier) We should update this to match the new signature

functorch/_src/make_functional.py
```diff
@@ -353,15 +359,28 @@ def compute_loss(params, x, t):
...
     if len(buffers) > 0:
         raise RuntimeError('make_functional(model): `model` has buffers. Please use '
                            'make_functional_with_buffers(model) instead.')
-    return FunctionalModule._create_from(model)
+    return FunctionalModule._create_from(model, disable_autograd_tracking=disable_autograd_tracking)
 
 
-def make_functional_with_buffers(model: nn.Module):
+def make_functional_with_buffers(model: nn.Module, disable_autograd_tracking: bool = False):
     """make_functional_with_buffers(model) -> func, params, buffers
```
samdow, line 384: nit: ditto here

samdow (approved):
LGTM! Just had the small nit that I missed earlier. Thanks for the patience with the doc + naming comments 😄

functorch/_src/make_functional.py
vfdev-5: @samdow sorry for delay. Actually, I didn't change this part of docstring as `disable_autograd_tracking` arg is not essential and IMO `make_functional(model) -> func, params` is showing essential usage.
What do you think ? I can change it though if you confirm. Thanks

vfdev-5: Oh, I see, I haven't remarked that previously. Thanks! Let me update the docstring here as well

samdow:
Merging since all of the failures look to be from upstream. Thanks for all your work on this @vfdev-5!

PR #718, opened 2022-04-20
Fix test following tls update in core

albanD:
This test now works after https://github.com/pytorch/pytorch/pull/75827


samdow:
Change looks to be now on nightly, thanks for the patch! Merged in but we still will see one more test failure (fix incoming)

PR #729, opened 2022-04-21
Add mode to AOTAutograd

samdow:
Using a mode, AOTAutograd is able to to trace factory functions. Because the tracer is saved as state on the PythonKey, this is doing some hacky stuff using globals to make the tracing work when we only have access to the PythonKey type (which is a requirement that will be lifted [here](https://github.com/pytorch/pytorch/pull/75965))

This also adds the ability to trace as a flag passed to make_fx...not sure if that's the API we want or if we want it to be always on


samdow:
Closing, will add upstream

PR #738, opened 2022-04-22
Make printing work again for PythonTensor

ezyang:


Chillee (approved):
Hell yeah, I actually have this locally but let's get it merged.

PR #739, opened 2022-04-22
Promote AOTAutograd decompositions to the main set

ezyang:
I want there to only be one set of decomps, so get rid of the out of line AOTAutograd decomps and clean them up.

### DIFF
--- a/functorch/_src/aot_autograd.py
+++ b/functorch/_src/aot_autograd.py
@@ -99,19 +99,6 @@ def normalize_as_list(x):
     return [x]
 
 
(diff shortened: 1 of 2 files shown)


Chillee:
Oh, these decompositions are meant to be "hacky"/"not necessarily correct" decompositions.

For example, previously I put a "decompositions" from `aten::_s_where` to `aten::where` here.

ezyang:
New plan on record is to make it easy for uses of tracers to both say what decomps they want, as well as define behavior by their own decomps. However, I do think these two decomps morally should be in our decompositions set.

PR #740, opened 2022-04-22
Add trace decomposition, invoke from C++

eellison:
TODO - integrate registrations of the decomposition w/jit into the `decompositions.py` file

Relies on https://github.com/pytorch/pytorch/pull/76252

cc @zou3519 @Chillee


functorch/_src/vmap.py
```diff
@@ -36,6 +37,16 @@ def _odict_unflatten(values, context):
 _register_pytree_node(OrderedDict, _odict_flatten, _odict_unflatten)
 
 
+def register_jit_decomposition(decomp):
+    assert decomp in decomposition_table, f"could not find {decomp}"
+    decomp_fn = decomposition_table[torch.ops.aten.trace.default]
```
eellison, line 42: oop

functorch/_src/decompositions.py
```diff
@@ -197,6 +197,11 @@ def silu_backward(grad_output: Tensor, self: Tensor) -> Tensor:
     return grad_output * sigmoid * (1 + self * (1 - sigmoid))
 
 
+@register_decomposition(aten.trace)
+def trace_decomp(x):
```
Chillee, line 201: nit: just rename this to trace

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -152,7 +153,10 @@ std::tuple<Tensor,optional<int64_t>> _unsafe_view_batch_rule(
 }
 
 Tensor trace_decomp(const Tensor& self) {
```
Chillee, line 155: can we make this a macro? or perhaps a boxed fallback?

Chillee (approved):
Looks good - might be nicer for the decomp stuff to be a boxed fallback instead.

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -152,7 +153,10 @@ std::tuple<Tensor,optional<int64_t>> _unsafe_view_batch_rule(
...
-  return at::sum(at::diagonal(self));
+  static auto * trace_exec = torch::jit::GetDecompositionExecutor("aten::trace(Tensor self) -> Tensor");
```
eellison, line 156: cc @ezyang do you know if there's any way to easily blug in boxed fallback here ?

eellison: nvm, @horace pointed me to https://github.com/pytorch/functorch/blob/main/functorch/csrc/BatchRulesHelper.h#L114

eellison: Actually... pointers still helpful

functorch/csrc/BatchRulesViews.cpp
eellison: Done... depends on https://github.com/pytorch/pytorch/pull/76493

zou3519 (approved):
This is super cool! I can't wait for this to be landed; my plan on record previously was to transcribe everything in decompositions.py into C++ that I needed

ezyang:
@zou3519 couldn't you have also just called into the Python directly

zou3519:
@ezyang, we have three short-term options really:
1. Transcribe the Python into C++. This is what I'm doing right now (e.g., in https://github.com/pytorch/functorch/pull/764) because I didn't want to think too hard
2. Call directly from C++ into Python. Requires us to either use Anjali's Python Op registration logic or implement something similar (a Python function that can be called from C++)
3. Use this PR (which is getting the decompositions into TorchScript and then being able to call the TorchScript from C++).

There's an open question around what the medium and long term state of this should be. That I am not sure about.
- for medium-term: It seems more lightweight to call into Python directly, but TorchScript is potentially faster due to not needing to get back into Python interpreter. Need to benchmark to see if this matters at all.
- for long-term: ???

ezyang:
I think the appropriate long term attitude is that decomps get traced through in a JIT (e.g., AOTAutograd via torchdynamo) and therefore it doesn't matter that they directly get run in Python.

PR #743, opened 2022-04-25
Generalize get_decompositions handling for OpOverloadPackets

ezyang:
Previously, it required only a single overload be registered to
a packet, if you asked for that packet in decompositions.  But
this is not really necessary; you could easily just return decomps
for all overloads in the packet.  This is strictly more general
and I can't imagine a situation where it's not what you want.


Chillee:
That's true - I think this was lost in the shuffle when __torch_dispatch__ switched to overloads.

PR #750, opened 2022-04-26
Minor improvements for _autograd_grad

ezyang:
I was really annoyed at the fact that we preallocate result
tensors for everything and then throw most of them out.  New
code variant doesn't do that.


zou3519 (approved):
Oh man, thanks for reading through the bad code I wrote, figuring out what it does, and improving it

zou3519:
Still applies cleanly and I re-tested this locally; merging

PR #753, opened 2022-04-26
L1 loss backward decomposition

samdow:
Adds decomposition for L1 loss backward

If self and target are of different sizes, it will warn (but not error) in the forward pass. I didn't add this to the backward call since it assumes the user will have seen the warning once during the forward

### DIFF
--- a/functorch/_src/decompositions.py
+++ b/functorch/_src/decompositions.py
@@ -535,6 +535,14 @@ def addmm(self: Tensor, mat1: Tensor, mat2: Tensor, beta: int = 1, alpha: int = 
     return beta * self + out
 
 
+@register_decomposition(aten.l1_loss_backward)
+def l1_loss_backward(grad_output: Tensor, self: Tensor, target: Tensor, reduction: int = Reduction.MEAN.value):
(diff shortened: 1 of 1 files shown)


functorch/_src/decompositions.py
```diff
@@ -489,6 +491,16 @@ def addmm(self: Tensor, mat1: Tensor, mat2: Tensor, beta: int = 1, alpha: int =
     return beta * self + out
 
 
+@register_decomposition(aten.l1_loss_backward)
+def l1_loss_backward(grad_output: Tensor, self: Tensor, target: Tensor, reduction: int = Reduction.MEAN.value):
+    assert(reduction in tuple(red.value for red in Reduction))
+    sign = torch.sign(self - target)
+
+    # numel doesn't work as the denominator if self is non-contiguous
```
zou3519, line 499: Why not?

samdow: Oh interesting. I thought I was seeing different results from playing around with it in Python but I think I must have messed something up locally. Switching back

zou3519 (approved):

PR #755, opened 2022-04-27
NLL Loss backward decomposition

samdow:
Error checks are pulled from [this function](https://github.com/pytorch/pytorch/blob/a079cbca0a0e312017665efa5fae00f24c7305ef/aten/src/ATen/native/LossNLL.cpp#L62)

Decomposition is pulled from batching rule

### DIFF
--- a/functorch/_src/decompositions.py
+++ b/functorch/_src/decompositions.py
@@ -285,6 +285,59 @@ def huber_loss_backward(grad_output: Tensor, self: Tensor, target: Tensor, reduc
     )
 
 
+@register_decomposition(aten.nll_loss_backward)
+def nll_loss_backward(grad_output: Tensor, self: Tensor, target: Tensor, weight: Optional[Tensor], reduction: int, ignore_index: int, total_weight: Tensor) -> Tensor:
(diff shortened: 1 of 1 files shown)


functorch/_src/decompositions.py
```diff
@@ -285,6 +285,55 @@ def huber_loss_backward(grad_output: Tensor, self: Tensor, target: Tensor, reduc
     )
 
 
+@register_decomposition(aten.nll_loss_backward)
+def nll_loss_backward(grad_output: Tensor, self: Tensor, target: Tensor, weight: Optional[Tensor], reduction: int, ignore_index: int, total_weight: Tensor) -> Tensor:
+    if self.dim() < 0 or self.dim() > 2:
```
Chillee, line 290: maybe just makes these asserts?

Chillee: In the future we might want to trace these into graphs, and asserts are easier to trace than control flow + raise errors :)

samdow: Changed

zou3519 (approved):
I'm happy as long as the tests pass :). Didn't think too hard about this but it looks correct.

PR #764, opened 2022-04-28
Use nll_loss_backward decomposition for jvp transform

zou3519:


zou3519:
Before anyone asks: this is a pretty big hack. Ideally we would register the decomposition with PyTorch's forward-mode AD and since functorch leverages the PyTorch forward-mode AD implementation and rules for its jvp transform, then we would get them for free.

Since this was relatively easy to do and I want to do some initial testing for this approach, I added this hack, but in parallel we should make progress toward being able to use decompositions with the PyTorch forward-mode AD subsystem.

PR #771, opened 2022-05-02
remove decompositions from functorch (but keep shims)

Chillee:
Removing decompositions since they've been ported into PyTorch core.


ezyang (approved):
thx

Chillee:
Sadly, Pytorch nightly has not updated yet >:(

PR #772, opened 2022-05-02
Port jvp decompositions to unify on python one

Chillee:
cc: @zou3519  @eellison 

ezpz


functorch/csrc/DynamicLayer.cpp
```diff
@@ -550,7 +551,7 @@ TORCH_LIBRARY_IMPL(_, FT_DYNAMIC_LAYER_BACK_MODE_KEY, m) {
   m.impl(#op "." #overload, torch::CppFunction::makeFromBoxedFunction<&dynamicLayerFrontFallbackWithOpId<ATenOpId::op ## _ ## overload>>());
 
 TORCH_LIBRARY_IMPL(aten, FT_DYNAMIC_LAYER_FRONT_MODE_KEY, m) {
-  FALLBACK_WITH_ID(nll_loss_backward);
+  RUN_JIT_DECOMPOSITION(nll_loss_backward);
```
zou3519, line 554: This decomposes nll_loss_backward for ALL transforms, which is not what we want: the grad transform by itself should still call nll_loss_backward

PR #774, opened 2022-05-03
Added chunks arg to vmap

vfdev-5:
### Description:
- Added chunks arg to vmap
- Added a test
- TODO: Update docstring

I had to use `del vars` in order to reduce total memory usage, for example while concatenating output chunks.

### Use-case: avoid OOM on cuda

Here is a code that fails with OOM on 11GB GPU without using chunks API but can compute the result with using 4 chunks:
```python
import torch
import torch.nn.functional as F
from torchvision.models import resnet101
import functorch as ft
from functorch.experimental import replace_all_batch_norm_modules_

device = "cuda"
model = resnet101(num_classes=10).eval().to(device)
replace_all_batch_norm_modules_(model)

fn, params, buffers = ft.make_functional_with_buffers(model, disable_autograd_tracking=True)

def compute_loss(params_, buffers_, x_, y_true_):
    y_pred = fn(params_, buffers_, x_.unsqueeze(0))
    loss = F.cross_entropy(y_pred, y_true_.unsqueeze(0))
    return loss , {"loss": loss, "y_pred": y_pred.squeeze(0)}

N = 32
x = torch.rand(N, 3, 244, 244, device=device)
y = torch.randint(0, 10, size=(N,), device=device)

has_aux = True
grad_fn = ft.grad(compute_loss, has_aux=has_aux)
chunks = 4
out = ft.vmap(grad_fn, in_dims=(None, None, 0, 0), chunks=chunks)(params, buffers, x, y)

print(len(out))
print(out[0][0].shape)
print(out[0][1].shape)
print(out[1]["loss"].shape)
print(out[1]["y_pred"].shape)
```
Output:
```
2
torch.Size([32, 64, 3, 7, 7])
torch.Size([32, 64])
torch.Size([32])
torch.Size([32, 10])
```

In the example, we can not test `chunks` = 2 or 3 as it still OOMs in vmap for the second or third chunk in autograd method.

### Benchmarks

<details>
<summary>
Benchmark code
</summary>

```python
import torch
import torch.nn.functional as F
from torchvision.models import resnet101
import functorch as ft
from functorch.experimental import replace_all_batch_norm_modules_

import torch.utils.benchmark as benchmark

device = "cuda"
model = resnet101(num_classes=10).eval().to(device)
replace_all_batch_norm_modules_(model)

fn, params, buffers = ft.make_functional_with_buffers(model, disable_autograd_tracking=True)

def compute_loss(params_, buffers_, x_, y_true_):
    y_pred = fn(params_, buffers_, x_.unsqueeze(0))
    loss = F.cross_entropy(y_pred, y_true_.unsqueeze(0))
    return loss , {"loss": loss, "y_pred": y_pred.squeeze(0)}

N = 32
x = torch.rand(N, 3, 244, 244, device=device)
y = torch.randint(0, 10, size=(N,), device=device)

has_aux = True
grad_fn = ft.grad(compute_loss, has_aux=has_aux)

# Compare takes a list of measurements which we'll save in results.
results = []

for chunks in [4, 8, 12, 16, 32]:
    label = 'Vmap with chunks'
    sub_label = f'chunks={chunks}'
    results.append(benchmark.Timer(
        stmt='vmap(grad_fn, in_dims=(None, None, 0, 0), chunks=chunks)(params, buffers, x, y)',
        globals={
            'vmap': ft.vmap,
            'grad_fn': grad_fn,
            'chunks': chunks,
            'params': params,
            'buffers': buffers,
            'x': x,
            'y': y
        },
        label=label,
        sub_label=sub_label,
        description='vmap with chunks',
    ).blocked_autorange(min_run_time=2))

compare = benchmark.Compare(results)
compare.print()
```

</details>

```
[-------- Vmap with chunks --------]
                 |  vmap with chunks
1 threads: -------------------------
      chunks=4   |        665.1     
      chunks=8   |        878.2     
      chunks=12  |       1186.7     
      chunks=16  |       1652.2     
      chunks=32  |       3528.1     

Times are in milliseconds (ms).
```

#### Memory usage checks with N=16, chunks=1 vs chunks=4

Results are measured with `torch.cuda.memory_allocated()`:

chunks | output CUDA memory (MB) | Peak memory usage
--- | --- | ---
1  | 2934 | -
4 | 2942 | 2948

### Limitations

- small number of chunks can not always help to prevent OOM on CUDA
- while concatenating output results we have a phase when chunks and concat output both exists

### DIFF
--- a/functorch/_src/vmap.py
+++ b/functorch/_src/vmap.py
@@ -205,7 +205,8 @@ def vmap(
         func: Callable,
         in_dims: in_dims_t = 0,
         out_dims: out_dims_t = 0,
-        randomness: str = 'error') -> Callable:
+        randomness: str = 'error',
+        chunks=1) -> Callable:
     """
     vmap is the vectorizing map; ``vmap(func)`` returns a new function that
     maps :attr:`func` over some dimension of the inputs. Semantically, vmap
@@ -235,6 +236,7 @@ def vmap(
             random functions will error. Default: 'error'. WARNING: this flag
             only applies to random PyTorch operations and does not apply to
             Python's random module or numpy randomness.
+        chunks (int): TODO
 
     Returns:
         Returns a new "batched" function. It takes the same inputs as
@@ -355,10 +357,22 @@ def vmap(
     if randomness not in ['error', 'different', 'same']:
         raise RuntimeError(f"Only allowed values for randomness are 'error', 'different', or 'same'. Got {randomness}")
 
-    @functools.wraps(func)
-    def wrapped(*args, **kwargs):
-        _check_out_dims_is_int_or_int_pytree(out_dims, func)
-        batch_size, flat_in_dims, flat_args, args_spec = _process_batched_inputs(in_dims, args, func)
+    if chunks == 1:
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
+            _check_out_dims_is_int_or_int_pytree(out_dims, func)
+            batch_size, flat_in_dims, flat_args, args_spec = _process_batched_inputs(in_dims, args, func)
+            vmap_level = _vmap_increment_nesting(batch_size, randomness)
+            try:
+                batched_inputs = _create_batched_inputs(flat_in_dims, flat_args, vmap_level, args_spec)
+                batched_outputs = func(*batched_inputs, **kwargs)
+                return _unwrap_batched(batched_outputs, out_dims, vmap_level, batch_size, func)
+            finally:
+                _vmap_decrement_nesting()
+
+        return wrapped
+
+    def _flat_vmap(batch_size, flat_in_dims, flat_args, args_spec, **kwargs):
         vmap_level = _vmap_increment_nesting(batch_size, randomness)
         try:
             batched_inputs = _create_batched_inputs(flat_in_dims, flat_args, vmap_level, args_spec)
@@ -366,4 +380,63 @@ def vmap(
             return _unwrap_batched(batched_outputs, out_dims, vmap_level, batch_size, func)
         finally:
             _vmap_decrement_nesting()
-    return wrapped
+
+    def _get_chunk_flat_args(flat_args_, flat_in_dims_, chunks_):
+        flat_args_chunks = tuple(
+            t.chunk(chunks_, dim=in_dim) if in_dim is not None else [t, ] * chunks_
+            for t, in_dim in zip(flat_args_, flat_in_dims_)
+        )
+        # transpose chunk dim and flatten structure
+        # chunks_flat_args is a list of flatten args
+        chunks_flat_args = zip(*flat_args_chunks)
+        return chunks_flat_args
+
+    def _flatten_chunks_output(chunks_output_):
+        # chunks_output is a list of chunked outputs
+        # flatten chunked outputs:
+        flat_chunks_output = []
+        arg_spec_list = []
+        for output in chunks_output_:
+            flat_output, arg_specs = tree_flatten(output)
+            flat_chunks_output.append(flat_output)
+            arg_spec_list.append(arg_specs)
+
+        arg_spec = arg_spec_list[0]  # all specs should be the same
+        # transpose chunk dim and flatten structure
+        # flat_output_chunks is flat list of chunks
+        flat_output_chunks = list(zip(*flat_chunks_output))
+        return flat_output_chunks, arg_spec
+
+    @functools.wraps(func)
+    def wrapped_with_chunks(*args, **kwargs):
+        _check_out_dims_is_int_or_int_pytree(out_dims, func)
+        _, flat_in_dims, flat_args, args_spec = _process_batched_inputs(in_dims, args, func)
+        # Chunk flat arguments
+        chunks_flat_args = _get_chunk_flat_args(flat_args, flat_in_dims, chunks)
+
+        # Apply vmap on chunks
+        chunks_output = []
+        for flat_args in chunks_flat_args:
+            batch_size = _validate_and_get_batch_size(flat_in_dims, flat_args)
+            chunks_output.append(
+                _flat_vmap(batch_size, flat_in_dims, flat_args, args_spec, **kwargs)
+            )
+
+        flat_output_chunks, arg_spec = _flatten_chunks_output(chunks_output)
+        # Removing temporary variables helps to reduce memory usage on device like CUDA
+        del chunks_output
+
+        # concat chunks on out_dim
+        flat_out_dims = _broadcast_to_and_flatten(out_dims, arg_spec)
+        assert len(flat_out_dims) == len(flat_output_chunks)
+        flat_output = []
+        for out_dim in flat_out_dims:
+            flat_output.append(torch.cat(flat_output_chunks[0], dim=out_dim))
+            # release source data
+            del flat_output_chunks[0]
+        del flat_output_chunks
+
+        # finally unflatten the output
+        return tree_unflatten(flat_output, arg_spec)
+
+    return wrapped_with_chunks
--- a/test/test_vmap.py
+++ b/test/test_vmap.py
@@ -2716,14 +2716,20 @@ class TestVmapOperators(Namespace.TestVmapBase):
 
         self.assertTrue(torch.randn(()).dim() == 0)
 
-    @parametrize('op', [torch.cos, torch.sinh], name_fn=lambda f: f.__name__)
-    def test_foobar_parametrize(self, op):
-        pass
+    @parametrize('in_dim', [0, 1, 2])
+    @parametrize('out_dim', [0, 1, 2])
+    def test_vmap_chunks(self, in_dim, out_dim):
+        x = torch.randn(16, 16, 16)
 
-    @parametrize('op2', [torch.cos, torch.sinh], name_fn=lambda f: f.__name__)
-    @parametrize('op1', [torch.abs, torch.acos], name_fn=lambda f: f.__name__)
-    def test_parametrize_multiple(self, op1, op2):
-        pass
+        def f(x):
+            y = x.sin()
+            return y
+
+        expected = vmap(f, in_dims=in_dim, out_dims=out_dim)(x)
+
+        for chunks in [1, 2, 3, 4, 7, 10, 16]:
+            output = vmap(f, in_dims=in_dim, out_dims=out_dim, chunks=chunks)(x)
+            self.assertEqual(output, expected)
 
 
 instantiate_parametrized_tests(TestVmapOperators)
@@ -2902,10 +2908,6 @@ class TestVmapBatchedGradient(Namespace.TestVmapBase):
         x = _get_rand_no_zeros(2, 3, device=device, requires_grad=True)
         self._batched_grad_test(torch.log1p, (x,))
         self._batched_grad_grad_test(torch.log1p, (x,))
-
-    @parametrize('param', ['foo', 'bar'])
-    def test_param_device(self, device, param):
-        pass
 
     @allowVmapFallbackUsage
     def test_max(self, device):


functorch/_src/vmap.py
```diff
@@ -355,15 +357,86 @@ def vmap(
...
+        for out_dim in flat_out_dims:
+            flat_output.append(torch.cat(flat_output_chunks[0], dim=out_dim))
+            # release source data
+            del flat_output_chunks[0]
+        del flat_output_chunks
+
+        # finally unflatten the output
+        return tree_unflatten(flat_output, arg_spec)
+
+    return wrapped_with_chunks
```
zou3519, line 442: I didn't read very closely, but the approach looks good to me (and I will give a closer read after we figure out the other details).

To summarize what I think is going on here: vmap(..., chunks=N) does the following:
- chunks the inputs
- does vmap on each chunk to produce an output
- cats each output chunk to produce the original
and does not do anything special, like device offloading, right?

If so, I want to bikeshed the API a little more. I'm worried about two cases:
- It seems to me that people will end up calling library code (that uses vmap with chunks=1), but it turns out that the library code OOMs
- People call library code that calls vmap with chunks=5, because this was the most optimal number at the time. However, it's been 2 years and they now have a larger machine...

Then, there are two options:
1) the library code must thread a `chunks` argument around. The API that the user call has a `chunks` argument.
2) We provide a context-manager-based interface for chunks that can override the previously set chunk count

I don't know which one of these is better, but here's what I mean for Option 2.

## Proposal

We provide a `vmap_chunks(chunks=1, override=True)` context manager

The way to use it is:
```
with vmap_chunks(chunks=4):
  functorch.vmap(f)(x)

with vmap_chunks(chunks=4):
  functorch.hessian(f)(x)
```

The context managers are nestable. 
```
# The innermost one gets precedence
with vmap_chunks(chunks=4):
  with vmap_chunks(chunks=2):
    # uses 2 chunks
    functorch.hessian(f)(x)

# Unless the outer one specifies override
with vmap_chunks(chunks=4, override=True):
  with vmap_chunks(chunks=2):
    # uses 4 chunks
    functorch.hessian(f)(x)
```

What are your thoughts? @vfdev-5 , (also @samdow -- you've been working a lot with context managers recently)

vfdev-5: Thanks for the review @zou3519 !

As for the two options, I was thinking about the first option:

> the library code must thread a chunks argument around. The API that the user call has a chunks argument.

which means that user specifies `chunks` at each call like : 
```python
chunks = 4
out1 = functorch.hessian(f, chunks=chunks)(x)
out2 = functorch.jacrev(f, chunks=chunks)(x)
```

If we would like to avoid specifying it on every call and yet without making complex context managers, we can think of a global variable:
```python
from functorch.globals import set_num_chunks, get_num_chunks

set_num_chunks(4)
print(get_num_chunks())  # 4

out1 = functorch.hessian(f)(x)
out2 = functorch.jacrev(f)(x)
```

I think introducing `vmap_chunks(chunks=1, override=True)` context manager may bring a lot of unnecessary complexity to handle (multiple CMs, override/do not override etc). Please fix me if I'm missing something important for the API.

vfdev-5: I see the problem. I agree with you that if it is a nested call, option 2 (`jacfwd(jacrev(f), chunks=4)`) is a way to compute with chunks.

> Now, let's assume we did option 2, using the global variable. Then what exactly does the following do? Is it option 1 or option 2?

```python
set_num_chunks(4)
print(get_num_chunks())  # 4
out1 = functorch.hessian(f)(x)
# under the hood:
# return jacfwd(jacrev(f), chunks=get_num_chunks())
```

vfdev-5: > Maybe 1 should not be solved using the chunks API?

There are few workarounds here. a) We can say in the docs about chunks that hessian does chunk on cols (`num_cols_in_jacobian`), if this does not pass memory size, user can express manually the hessian as `jacfwd(jacrev(f, chunks=num_rows_in_jacobian), chunks=num_cols_in_jacobian)` and tune the params such that it could fit the memory.
b) try/except in hessian with  `num_cols_in_jacobian` only option and if fails with OOM try `jacfwd(jacrev(f, chunks=chunks), chunks=chunks)`. I'm not fan of such automagic options.
c) exceptionally for hessian we can have 2 chunk args `num_rows_in_jacobian`, `num_cols_in_jacobian`.

In `jacrev` there is also an expensive `_construct_standard_basis_for` which may be tried to to be done differently...

> "I am running into OOM and I want to tune a some numbers until my program stops OOM-ing)"

This looks like "auto" batch size search for NN. This can be an option.

test/test_eager_transforms.py
```diff
@@ -2946,7 +2946,6 @@ def f(x: torch.Tensor) -> torch.Tensor:
             return x
         self._check_functionalize_correctness(f, torch.zeros(4, 2, device=device))
 
-
```
vfdev-5, line 2949: My flake8 was failing on these files... Maybe different version from the one in lintrunner

functorch/_src/vmap.py
zou3519: Punting this discussion to the future, per offline discussion

functorch/_src/vmap.py
```diff
@@ -366,4 +366,119 @@ def wrapped(*args, **kwargs):
...
+
+    def _get_chunk_flat_args(flat_args_, flat_in_dims_, chunks_):
+        flat_args_chunks = tuple(
+            t.chunk(chunks_, dim=in_dim) if in_dim is not None else [t, ] * chunks_
+            for t, in_dim in zip(flat_args_, flat_in_dims_)
+        )
+        # transpose chunk dim and flatten structure
+        # chunks_flat_args is a list of flatten args
+        chunks_flat_args = zip(*flat_args_chunks)
+        return chunks_flat_args
```
zou3519, line 434: I feel like we don't actually need to peek into vmap internals to do this, but I also have not thought about this problem as much as you have. 

Could we implement chunk_vmap like the following?
```py
def chunk_vmap(func, in_dims, out_dims, randomness, chunks):
  if chunks == 1:
    return vmap(func, in_dims, out_dims, randomness)
  def inner(*args):
    chunked_args = chunk_args(args)
    chunked_outputs = [vmap(func, in_dims, out_dims, randomness)(args) for args in chunked_args]
    outputs = concat_chunks(chunked_outputs)
    return outputs
  return inner
```
This way if someone changes how vmap works (e.g. adds a new randomness option), we do not have to change it inside vmap and here

functorch/_src/vmap.py
```diff
@@ -366,4 +366,119 @@ def wrapped(*args, **kwargs):
...
+
+    Returns:
+        Returns a new "batched" function. It takes the same inputs as
+        :attr:`func`, except each input has an extra dimension at the index
+        specified by :attr:`in_dims`. It takes returns the same outputs as
+        :attr:`func`, except each output has an extra dimension at the index
+        specified by :attr:`out_dims`.
+    """
+    if randomness not in ['error', 'different', 'same']:
+        raise RuntimeError(f"Only allowed values for randomness are 'error', 'different', or 'same'. Got {randomness}")
```
zou3519, line 412: I just realized: randomness="same" doesn't actually work for chunked vmap, right? cc @samdow 

Let's assume we're doing the following:
```
def f(x):
  return torch.rand([])
  
x = torch.ones(4)

# returns the same value
vmap(f, randomness='same')(x)

# Returns torch.tensor([123, 123, 456, 456]). Read on for explanation.
chunk_vmap(f, randomness='same')(x)

# As implemented in this PR, chunk_vmap essentially calls the following:
chunk0 = vmap(f, randomness='same')(x[:2])
chunk1 = vmap(f, randomness='same')(x[2:])
result = torch.stack(chunk0, chunk1)
```

If we agree that is what this PR is doing, then one possible thing we can do right now:
- since we're tossing chunk_vmap in functorch.experimental.chunk_vmap, it's OK for this to not work and we can keep on thinking about what the semantics should be
- therefore it is OK to say we do not support randomness = same with chunk_vmap right now

But tbh I am not sure how we will support randomness = same with chunk_vmap in the future.

functorch/_src/vmap.py
```diff
@@ -366,4 +366,119 @@ def wrapped(*args, **kwargs):
...
+            it should have one element per output. Default: 0.
+        randomness (str): Specifies whether the randomness in this
+            vmap should be the same or different across batches. If 'different',
+            the randomness for each batch will be different. If 'same', the
+            randomness will be the same across batches. If 'error', any calls to
+            random functions will error. Default: 'error'. WARNING: this flag
+            only applies to random PyTorch operations and does not apply to
+            Python's random module or numpy randomness.
+        chunks (int): Number of chunks to use to split the input data. Default is 2.
+            If equals to 1 then :func:`vjp` is called.
```
zou3519, line 402: nit: vjp -> vmap

functorch/_src/vmap.py
vfdev-5: Right now the implementation is almost like you described but uses flat_vmap instead of vmap as it is more simple to chunk flat args.
If doing as you suggest:
```python
    chunked_args = chunk_args(args)
    chunked_outputs = [vmap(func, in_dims, out_dims, randomness)(args) for args in chunked_args]
```
Behind the scene, we may probably flatten and unflatten 2 times more the args as `chunk_args` will flatten, chunk, unflatten and vmap will also flatten, flat_vmap, unflatten.

If flatten/unflatten is not a problem we can try to code it according to you suggestion.

functorch/_src/vmap.py
vfdev-5: For same we could maybe do something like:
```
        chunks_output = []
        rs = torch.get_rng_state()
        for flat_args in chunks_flat_args:
            batch_size = _validate_and_get_batch_size(flat_in_dims, flat_args)
            torch.set_rng_state(rs)
            chunks_output.append(
                _flat_vmap(batch_size, flat_in_dims, flat_args, args_spec, **kwargs)
            )

```

functorch/_src/vmap.py
zou3519: > If flatten/unflatten is not a problem

I see, thanks for the clarification. I agree, it would be nice to avoid too many flattens and unflattens. Let's stick with your approach; I will add some suggestions so that we repeat ourselves less.

functorch/_src/vmap.py
zou3519, line 412: Can we refactor this into a `check_randomness` helper function that is used both here and in vmap?

functorch/_src/vmap.py
```diff
@@ -366,4 +366,119 @@ def wrapped(*args, **kwargs):
...
+
+    def _flat_vmap(batch_size, flat_in_dims, flat_args, args_spec, **kwargs):
+        vmap_level = _vmap_increment_nesting(batch_size, randomness)
+        try:
+            batched_inputs = _create_batched_inputs(flat_in_dims, flat_args, vmap_level, args_spec)
+            batched_outputs = func(*batched_inputs, **kwargs)
+            return _unwrap_batched(batched_outputs, out_dims, vmap_level, batch_size, func)
+        finally:
+            _vmap_decrement_nesting()
+
```
zou3519, line 425: Can we (should we?) refactor both vmap and chunk_vmap to call `_flat_vmap`?

functorch/_src/vmap.py
zou3519: Does this require the dimension being vmapped over to be evenly divisible by chunks?

zou3519 (commented):
Code mostly LGTM, thanks for explaining the approach, @vfdev-5. I had some suggestions to put things into helper functions (so we can share things between the chunk_vmap and vmap implementations), and we should figure out what to do with randomness="same".

functorch/_src/vmap.py
zou3519: I thought about this some more. The approach here works!

> Do you think this can depend on dim values ?

In general, randomness depends on how many random numbers you're generating. So if we have the following:
```
def f(N):
  torch.manual_seed(0)
  _ = torch.randn(N)
  return torch.randn([])
```
Then `f(0)` and `f(1)` return different things. Each random number that is generated advance the global RNG state forward by one.

## The case I was concerned about

```py
def f(x):
  torch.randn_like(x)
  return torch.randn([])

x = torch.randn(3)
torch.manual_seed(0)
expected = vmap(f, randomness='same')(x)
print(expected)

x0, x1 = x.chunk(2)
torch.manual_seed(0)
y0 = vmap(f, randomness='same')(x0)
torch.manual_seed(0)
y1 = vmap(f, randomness='same')(x1)
result = torch.cat([y0, y1], 0)
print(result)
```
it turns out that when `randomness='same'`, then we end up generating the same amount of random numbers for the `randn_like` call (despite the bdims having different shapes), so the final torch.randn call always returns the same thing.

## In conclusion

Let's go with your approach. Please try to add additional tests for it with some examples like the one above)

test/test_vmap.py
```diff
@@ -2719,14 +2719,31 @@ def naive_f(x, shape):
...
 
-    @parametrize('op', [torch.cos, torch.sinh], name_fn=lambda f: f.__name__)
-    def test_foobar_parametrize(self, op):
-        pass
+    @parametrize('in_dim', [0, 1, 2])
+    @parametrize('out_dim', [0, 1, 2])
+    @parametrize('randomness', ['error', 'different', 'same'])
+    def test_chunk_vmap(self, in_dim, out_dim, randomness):
+
+        if (in_dim > 0 or out_dim > 0) and randomness == "different":
```
vfdev-5, line 2727: I have an impression that vmap with randomness="different" and in_dim > 0 is not working as expected:
```python
in_dim = 1
out_dim = 0
randomness = "different"

torch.manual_seed(1)
x = torch.randn(4, 5)

def f(x):
    y = x.sin() + torch.rand_like(x)
    return y

rs = torch.get_rng_state()

expected = ft.vmap(f, in_dims=in_dim, out_dims=out_dim, randomness=randomness)(x)
print("expected:", expected)

xs = x.split(1, dim=in_dim)
torch.set_rng_state(rs)

out = []
for x in xs:
    out.append(ft.vmap(f, in_dims=in_dim, out_dims=out_dim, randomness=randomness)(x))
result = torch.cat(out, out_dim)
print("output:", result)
```
Output:
```
expected: tensor([[-0.0599,  0.5304,  1.2153,  0.1304],
        [-0.2651,  0.2595,  0.5518,  0.0550],
        [ 0.1056,  1.2283,  0.4167,  0.7079],
        [-0.7316,  0.6618, -0.1340,  1.2699],
        [ 1.7522,  1.6782, -0.6967,  1.8902]])
output: tensor([[-0.0599,  0.6587,  1.6947,  0.1364],
        [ 0.3088, -0.3270,  0.7797, -0.2103],
        [-0.3718,  1.4794,  0.2957,  0.9765],
        [-0.6436,  0.8703, -0.5599,  0.5335],
        [ 1.5329,  1.2997,  0.2820,  1.8902]])
```

However, results are the same if `in_dim=0`. Maybe, i should miss something...

cc @samdow

zou3519: This is expected I think. The batch rule for randn_like moves the batch dim to the front before calling randn_like: https://github.com/pytorch/functorch/blob/fb70a3cb588c4b937bb21479de762179b61b7661/functorch/csrc/BatchRulesRandomness.cpp#L164 . So the tensor that gets generated by using a shape of (N, 5) is different from the numbers that get generated by using a shape of (5, N) (the numbers two tensors will be transposed of each other).

In general, vmap(f) with randomness="different" does not promise the same semantics as doing a for-loop over f because the operations get executed in a different sequence, thereby advancing the global RNG state differently

zou3519 (commented):
I'm convinced the behavior you're seeing isn't actually a problem. I added a suggestion for how to test that randomness='different' is doing what we expect it to be doing (it should be giving us slices that are all unique). After that, this should be good to go

functorch/_src/vmap.py
```diff
@@ -352,18 +352,139 @@ def vmap(
...
+def chunk_vmap(
+        func: Callable,
+        in_dims: in_dims_t = 0,
+        out_dims: out_dims_t = 0,
+        randomness: str = 'error',
+        chunks=2) -> Callable:
+    """
+    chunk_vmap is the vectorizing map using chunks of input data. Splitting data into chunks
+    can help to prevent out-of-memory issues. For more details about vectorizing map, see :func:`vmap`.
```
zou3519, line 376: nit: (I can also write this later if you don't have bandwidth). It might be easier to explain `chunk_vmap` as a mix between vmap (which vectorizes everything) and map (which executes things sequentially). `chunk_vmap` vectorize things `#chunks` at a time.

zou3519 (approved):

PR #777, opened 2022-05-04
Fix advanced indexing

### LINKED ISSUE #773: Nested advanced indexing returns dimensions in the wrong order
```
from functorch import vmap
import torch
e = torch.rand(7, 4)
idx = torch.LongTensor([0, 1, 2, 3, 4, 5]).view(2, 3, 1)

# simple reference implementation for comparison
def _fake_vmap(f, in_dims=0):
    def w(input):
        r = [f(input.select(in_dims, i)) for i in range(input.size(in_dims))]
        return torch.stack(r, in_dims)
    return w

def with_vmap(_vmap):
    def g(idx_):
        def f(e_):
            return e_[idx_]
        return _vmap(f, in_dims=1)(e)
    r = _vmap(_vmap(g))(idx)
    return r

a = with_vmap(vmap)
b = with_vmap(_fake_vmap)
print(a.size(), b.size())
assert a == b
```

Produces:
```
torch.Size([4, 2, 3, 1]) torch.Size([2, 3, 1, 4])
Traceback (most recent call last):
  File "v.py", line 25, in <module>
    assert a == b
RuntimeError: The size of tensor a (2) must match the size of tensor b (3) at non-singleton dimension 1
```

samdow:
Fixes #773

Basically what was happening was that when we were only batching on the input tensor, we added None to the start so it became `x[idx]` -> `x[:, idx]`. This is fine until you start nesting vmaps that batch idx where, as the issue found, the batches get misaligned (note that the fake_vmap shape in the issue is also a little off, we default out_dim=0 instead of out_dim=in_dim).

It was tricky to add tests with nested vmaps using the current harness, so I added tests with empty indexing, which can get us the same misaligned behavior as well as the test provided in the issue (with out_dims added to get matching behavior)

### DIFF
--- a/functorch/csrc/BatchRulesScatterOps.cpp
+++ b/functorch/csrc/BatchRulesScatterOps.cpp
@@ -29,29 +29,38 @@ std::tuple<std::vector<optional<Tensor>>, int64_t, int64_t> batchIndices(
   //
   // 2. self is not batched, some indices are batched.
   // In this case, we don't need to do anything - indices will automatically
-  // broadcast to work with the unbatched self.
+  // broadcast to work with the unbatched self. If there are leading Nones or
+  // empty tensors in the indices, the batch dimension will appear after
+  // those leading blanks, so the 1st ret value captures that for the batching
+  // rules that use this
   //
   // 3. self is batched, some indices are batched.
   // In this case, we simply need to add an arange that indexes along the first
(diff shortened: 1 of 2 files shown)


Chillee:
:O nice! I took a look at it but couldn't figure it out.

PR #778, opened 2022-05-04
Try adding log_sigmoid_forward decomp

Chillee:
Doesn't work, gonna leave this here.

```
import torch
import torch.nn.functional as F
from torch import tensor
from functorch import jvp
aten = torch.ops.aten

def f(x):
    out = F.logsigmoid(x)
    print(out)
    return out

print(jvp(f, (torch.randn(5),), (torch.randn(5),)))
```


eellison:
@samdow do you mind linking me the old richard hack ? How are they supposed to be returned ? as a std::tuple<Tensor, Tensor> instead of a c10::Tuple<Tensor, Tensor>?

samdow:
@eellison For sure! [Here's](https://github.com/pytorch/functorch/commit/a22d15ff8104774d8c42bb4542a7454f41bb9463) a version of it for log_sigmoid that works using his old framework.

Flagging what I'm pretty sure isn't working right now. In the fixed version, [here](https://github.com/pytorch/functorch/commit/a22d15ff8104774d8c42bb4542a7454f41bb9463#diff-8d75d5141cd5d6feb6d29db3e15bf67070f4f21f34df02e90fcc76a0c3511df1R421-R422) we are separately pushing each element of the tuple (instead of pushing the tuple as a single element). Based on the errors and traces I was seeing, I'm guessing that the JIT version is trying to push the tuple as a single element

eellison:
Can you do something like this ?
```
diff --git a/functorch/csrc/BatchRulesHelper.cpp b/functorch/csrc/BatchRulesHelper.cpp
index adf5c4b..3118a68 100644
--- a/functorch/csrc/BatchRulesHelper.cpp
+++ b/functorch/csrc/BatchRulesHelper.cpp
@@ -138,6 +138,13 @@ void run_jit_decomposition(const c10::OperatorHandle& op, torch::jit::Stack* sta
   // TODO: templatize based on op and keep static trace_exec
   auto * trace_exec = torch::jit::GetDecompositionExecutor(schema);
   trace_exec->run((*stack));
+  if (stack->back().isTuple()) {
+    IValue tup = stack->back();
+    stack->pop_back();
+    for (const auto& elem: tup.toTuple()->elements()) {
+      stack->push_back(elem);
+    }
+  }
 }
 
 }}
 ```

samdow:
Thanks @eellison! That totally worked for me locally. I'll expect we see similarly in CI

@Chillee I added the commit directly here so you can continue working on this. Feel free to kick that commit as you test out like running decomps on the autograd key.

Also if either of you wants to take the commit as a separate PR/commit on main, feel free to! I'm also happy to bundle it with log_sigmoid, don't really care

eellison:
you can bundle, thanks!

Chillee:
@samdow Will do - I can land this. Thanks!

PR #779, opened 2022-05-05
added list clearing FX codegen for backwards in AOTAutograd.

Chillee:


ngimel (approved):
Nice!

Chillee:
Note: I've currently not bothered to land this PR, as Torchscript seems to keep all of the input activations alive.

We've verified that this reduces memory usage as expected if the graph isn't compiled with Torchscript: https://github.com/albanD/subclass_zoo/pull/41

PR #783, opened 2022-05-06
add lintrunner support

suo:
Black and flake8 for now + workflow in CI

Your codebase is not clean for either, so either fix em all or define
exclusions.

You can run lint on all files by doing
```
git grep -Il . | xargs lintrunner
```

You can autoformat all files by doing
```
git grep -Il . | xargs lintrunner f
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,24 +7,52 @@ on:
   pull_request:
 
 jobs:
-  flake8-py3:
+  lintrunner:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@v2
       - name: Setup Python
         uses: actions/setup-python@v2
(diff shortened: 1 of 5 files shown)


suo:
@Chillee @zou3519: this is the result of running on all files (flake8 only, since black obviously has a billion changes):
https://gist.github.com/suo/e4b28db0f3139a069c00022e77dfb323.

This is primarily because you guys don't install any of the common plugins for flake8 (which you maybe should?). Options, in order of good idea to bad idea:
1. fix these lints
2. turn off the plugins and just run raw flake8
3. turn off lintrunner on master and slowly fix them

Chillee:
Makes sense, I'll fix flake8 + black issues.

Also cc: @vfdev-5 @samdow

PR #784, opened 2022-05-06
fbcode sync 2022-05-06 15:54:41

ZolotukhinM:


samdow (approved):

samdow:
Not sure why this didn't autoclose, so closing it by hand (so that we don't get confused by the open PR) and setting up an issue

PR #788, opened 2022-05-09
Update neural_tangent_kernels.ipynb

ain-soph:
Fix a small bug
```python3
    if compute == 'full':
        return result
    if compute == 'trace':
        return torch.einsum('NMKK->NM')        # should be torch.einsum('NMKK->NM', result)
    if compute == 'diagonal':
        return torch.einsum('NMKK->NMK')        # should be torch.einsum('NMKK->NMK', result)
```

### DIFF
--- a/notebooks/neural_tangent_kernels.ipynb
+++ b/notebooks/neural_tangent_kernels.ipynb
@@ -288,9 +288,9 @@
     "    if compute == 'full':\n",
     "        return result\n",
     "    if compute == 'trace':\n",
-    "        return torch.einsum('NMKK->NM')\n",
+    "        return torch.einsum('NMKK->NM', result)\n",
     "    if compute == 'diagonal':\n",
-    "        return torch.einsum('NMKK->NMK')"
+    "        return torch.einsum('NMKK->NMK', result)"
    ]
   },
   {


ain-soph:
I follow the tutorial and implement the version without using `functorch`. I wonder what's the advantage to use `functorch`?

```python3

def ntk(module: nn.Module, input1: torch.Tensor, input2: torch.Tensor,
        parameters: dict[str, nn.Parameter] = None,
        compute='full') -> list[torch.Tensor]:
    einsum_expr: str = ''
    match compute:
        case 'full':
            einsum_expr = 'Naf,Mbf->NMab'
        case 'trace':
            einsum_expr = 'Naf,Maf->NM'
        case 'diagonal':
            einsum_expr = 'Naf,Maf->NMa'
        case _:
            raise ValueError(compute)

    if parameters is None:
        parameters = dict(module.named_parameters())
    keys, values = zip(*parameters.items())

    def func(*params: torch.Tensor, _input: torch.Tensor = None):
        _output: torch.Tensor = _stateless.functional_call(
            module, {n: p for n, p in zip(keys, params)}, _input)
        return _output  # (N, C)

    jac1: tuple[torch.Tensor] = torch.autograd.functional.jacobian(
        functools.partial(func, _input=input1), values, vectorize=True)
    jac2: tuple[torch.Tensor] = torch.autograd.functional.jacobian(
        functools.partial(func, _input=input2), values, vectorize=True)
    jac1 = [j.flatten(2) for j in jac1]
    jac2 = [j.flatten(2) for j in jac2]
    ntk_list = [torch.einsum(einsum_expr, j1, j2) for j1, j2 in zip(jac1, jac2)]
    return ntk_list

```

Chillee:
Thanks!

PR #791, opened 2022-05-10
Add forward over reverse support for binary cross entropy

samdow:
Layer norm is having some numerical trouble so that one will be another moment


test/test_ops.py
```diff
@@ -1179,6 +1177,7 @@ def test_vjpvmap(self, device, dtype, op):
         xfail('scatter_reduce', 'mean'),
         xfail('scatter_reduce', 'prod'),
         skip('linalg.householder_product', '', device_type='cuda'),  # flaky, I'm not sure why
+        xfail('nn.functional.binary_cross_entropy_with_logits')
```
samdow, line 1180: I assumed this would work for both of them but it looked like this was having numerical issues (versus the not implemented issues of binary cross entropy). I can look into this later if it's something we really want to support

PR #797, opened 2022-05-12
Minor fix2

cyyever:
Some minor fixes found by static analyzer


samdow (approved):
LGTM! Thanks!

(failures are from upstream. The C++ is still able to compile, which should be the biggest test for most of these changes)

PR #798, opened 2022-05-12
[Draft] Jiterator as a compiler backend

SherlockNoMad:
codegen part of jiterator will be moved to torch.inductor.


SherlockNoMad:
PR moved to torchdynamo.

PR #802, opened 2022-05-13
Make fx tracing use input tensor copies to prevent side effects

byronyi:
Some side effects could possibly occur during `flat_fn` and `make_fx`, making unsound changes to the arguments (e.g. module buffers). 

Here I try to clone the arguments for compilation, so we always operate on user arguments with compiled function.

### DIFF
--- a/functorch/_src/aot_autograd.py
+++ b/functorch/_src/aot_autograd.py
@@ -147,8 +147,11 @@ def create_aot_autograd_function(
         def forward(ctx, *flat_tensor_args):
             nonlocal compiled_fw, compiled_bw, num_outs
             if compiled_fw is None:
+                flat_tensor_args_clone = pytree.tree_map(
+                    lambda x: x.clone().detach().requires_grad_(x.requires_grad)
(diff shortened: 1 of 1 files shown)


Chillee:
Hi, thanks for the PR :)

Most of the work on this tracer is now going on here: https://github.com/pytorch/pytorch/blob/master/torch/fx/experimental/proxy_tensor.py

However, about the specific proposal in this PR.... I don't know how I feel about it. I suppose in principle it makes sense, but if you're relying on inputs to your function to not be mutated in the first place then I feel like you might be using `make_fx` wrong.

byronyi:
A little bit more context: this issue hits me when I try to trace through batch norm in training mode, as the number of tracked batches is mutated during tracing. 

I agree with you: this seems more like a band-aid fix before the functionalization pass landed in core.

byronyi:
> if you're relying on inputs to your function to not be mutated in the first place then I feel like you might be using `make_fx` wrong.

Completely agree, but it seems we have to deal with this kind of cases when using it with dynamo.

PR #807, opened 2022-05-13
add layer norm support, clean up some binary cross entropy support

samdow:
This adds forward over reverse support for layer norm (it needs to be done a little hackily, some explanation is inline. But it has a slightly different decomposition for layer norm backward that shouldn't be used except for this temporary hack so that we can track the input gradients properly)

In doing so, it also generalized some of the binary cross entropy hacks for in_dims and found a small bug in there (the tolerance needed to be a little bigger) so there's some minor fixes around that

### DIFF
--- a/functorch/_src/decompositions.py
+++ b/functorch/_src/decompositions.py
@@ -1,7 +1,7 @@
 import torch
 from torch import Tensor
 import torch._decomp
-from typing import Tuple
+from typing import Tuple, List, Optional
 
 aten = torch.ops.aten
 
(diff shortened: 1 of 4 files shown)


Chillee:
> But it has a slightly different decomposition for layer norm backward that shouldn't be used except for this temporary hack so that we can track the input gradients properly

What's the issue here?

samdow:
> What's the issue here?

The saved mean and rstd don't propagate the gradients from the input (which is fair, they assume that they're using a handwritten kernel for backprop). In order to have the gradients propagate from the input through mean and rstd, we have to recompute them using input. This is hacky and gross

PR #810, opened 2022-05-17
Use Functionalization pass and guard with a global config flag

anijain2305:
PR for collaboration and hunting down bugs.


anijain2305:
cc @bdhirsh

PR #818, opened 2022-05-18
Add extremal testing for forward over reverse using decompositions 

samdow:
This adds tests that have tensors filled with -1000, 0, or 1000 (-1000 and 1000 because it hits inf for exp(-x) or exp(x), respectively) for the functions that use decompositions for forward over reverse coverage

~Currently, there's a bug in the nll loss decomposition that I didn't catch earlier so some of the shapes are commented out (specifically it fails if the weight is 0). I'm working on a fix for that before the branch cut tomorrow~ This bug was coming when weight was all 0s, so we decided that a user should not end up doing this. I'll check that this works with a weight of mostly 0s and a single 1 entry

### DIFF
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -1306,7 +1306,7 @@ def _register_jit_decomposition(decomp, use_python=False):
 
 
 _register_jit_decomposition(torch.ops.aten.trace.default)
-_register_jit_decomposition(torch.ops.aten.nll_loss_backward.default)
+_register_jit_decomposition(torch.ops.aten.nll_loss_backward.default, use_python=True)
 _register_jit_decomposition(torch.ops.aten.nll_loss2d_backward.default)
 _register_jit_decomposition(torch.ops.aten.mse_loss_backward.default)
 _register_jit_decomposition(torch.ops.aten.l1_loss_backward.default)
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -1306,7 +1306,7 @@ def _register_jit_decomposition(decomp, use_python=False):
 
 
 _register_jit_decomposition(torch.ops.aten.trace.default)
-_register_jit_decomposition(torch.ops.aten.nll_loss_backward.default, use_python=True)
+_register_jit_decomposition(torch.ops.aten.nll_loss_backward.default)
 _register_jit_decomposition(torch.ops.aten.nll_loss2d_backward.default)
 _register_jit_decomposition(torch.ops.aten.mse_loss_backward.default)
 _register_jit_decomposition(torch.ops.aten.l1_loss_backward.default)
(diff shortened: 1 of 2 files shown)


functorch/_src/eager_transforms.py
```diff
@@ -1306,7 +1306,7 @@ def get_function_def(sig):
 
 
 _register_jit_decomposition(torch.ops.aten.trace.default)
-_register_jit_decomposition(torch.ops.aten.nll_loss_backward.default)
+_register_jit_decomposition(torch.ops.aten.nll_loss_backward.default, use_python=True)
```
zou3519, line 1309: Why `use_python=True`?

test/test_ops.py
```diff
@@ -1102,6 +1104,27 @@ def test_vjpvmap(self, device, dtype, op):
 
                 self.assertEqual(result_vjps, expected_vjps)
 
+    def _compare_jacobians(self, fn, cotangents_and_primals, argnums=None, atol_rtol=None):
```
zou3519, line 1107: nit: `_compare_jacobians_of_vjp`? The original name was a bit misleading, that is my bad

test/test_ops.py
```diff
@@ -1102,6 +1104,27 @@ def test_vjpvmap(self, device, dtype, op):
...
+            argnums = tuple(range(len(cotangents_and_primals)))
+
+        def get_vjp(cotangents, *primals):
+            _, vjp_fn = vjp(fn, *primals)
+            return vjp_fn(cotangents)
+
+        jacobian_jvp = jacfwd(get_vjp, argnums)(*cotangents_and_primals)
+        jacobian_vjp = jacrev(get_vjp, argnums)(*cotangents_and_primals)
+
+        # For dtype changing operations, the jacobians have different dtype.
```
zou3519, line 1118: nit: would be nice to check that the jacobian is of the correct dtype, but I'm not sure if it is worth it. It seems complicated to do:

If the operation has inputs of dtype A and outputs of dtype B:
- then the vjp is a function that takes inputs of dtype (A, B) and gives outputs of dtype B
- so the jacobian_vjp should return jacobians of dtype A (for the primals) and dtype B (for the cotangents)
- the jacobian_jvp should return jacobians of dtype B

zou3519 (approved):
LGTM, some minor nits (feel free to ignore them)

functorch/_src/eager_transforms.py
samdow: Whoops yikes. From testing. Removed

test/test_ops.py
samdow: > would be nice to check that the jacobian is of the correct dtype

The other difficult thing is that none of the decompositions we're testing change dtype with the inputs we're using. In other words, currently all of the `jvpvjp` tests and these new tests pass without these lines. I'll file an issue that we should see if we can add some tests that change dtypes and see if we can update this section with that

PR #821, opened 2022-05-19
[test_jvpvjp] check if custom formula is available

samdow:
Adds a check to see if the forward over reverse tests where we are using a decomposition should actually be using a custom formula. This way, we can get signal to remove decompositions as the forwards over reverse support gets better

(tested with `nn.functional.linear`, failure does show up with error message: 
`AssertionError: True is not false : nn.functional.linear now supports forward over reverse without a decomposition`)

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1257,6 +1257,8 @@ class TestOperators(TestCase):
                 'nn.functional.layer_norm'
             }
             if op.name in FUNCTORCH_HAS_FORMULA_BUT_NOT_PYTORCH:
+                self.assertFalse(op.supports_fwgrad_bwgrad,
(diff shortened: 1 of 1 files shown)


test/test_ops.py
```diff
@@ -1257,6 +1257,8 @@ def get_vjp(cotangents, *primals):
                 'nn.functional.layer_norm'
             }
             if op.name in FUNCTORCH_HAS_FORMULA_BUT_NOT_PYTORCH:
+                self.assertFalse(op.supports_fwgrad_bwgrad,
+                                 f"{op.name} now supports forward over reverse without a decomposition")
```
zou3519, line 1261: nit: Add something like "we should remove the decomposition" so there are some instructions for people who are not you and me

zou3519 (approved):

PR #828, opened 2022-05-20
Add the ability to trace TensorList in-place ops

alanwaketan:
Summary:
To trace through c10d::all_gather, AOT needs to support TensorList
in-place ops.

Test Plan:
WIP.


byronyi:
This PR should be raised in core PyTorch as `make_fx` is moved there.

Issue #833, opened 2022-05-25
Replace `addr_decomp` with the one in PyTorch Core.


Chillee:
We have a decomposition for addr in functorch: https://github.com/pytorch/functorch/blob/aa3ca511900eb73ca4de32c6f7b182c01b9043b2/functorch/csrc/BatchRulesBinaryOps.cpp#L267

But we've also landed an `addr` decomposition in PyTorch core: https://github.com/pytorch/pytorch/pull/78014

We can remove the one in Functorch by leveraging the one in PyTorch core. 

1. Add it here: https://github.com/pytorch/functorch/blob/main/functorch/_src/eager_transforms.py#L1308
2. Modify the [addr registration](https://github.com/pytorch/functorch/blob/aa3ca511900eb73ca4de32c6f7b182c01b9043b2/functorch/csrc/BatchRulesBinaryOps.cpp#L267) to call into a decomposition instead (https://github.com/pytorch/functorch/blob/main/functorch/csrc/BatchRulesViews.cpp#L510).

Chillee:
Closed by https://github.com/pytorch/functorch/pull/836

PR #836, opened 2022-05-26
 Replace addr_decomp with the one in PyTorch Core, issue #833

yushangdi:
Remove the `addr` decomposition in `functorch` to use the one in  PyTorch core. 
1. removed the `addr` decomposition in functorch/csrc/BatchRulesBinaryOps.cpp
2. registered the Pytorch core decomposition in functorch/_src/eager_transforms.py. A new register method `_register_jit_decomposition_bypass_script` is implemented to bypass torch script. The existing `_register_jit_decomposition` does not work for `addr` because the Tensor types generated cannot be unioned by torch script.

### DIFF
--- a/functorch/_src/eager_transforms.py
+++ b/functorch/_src/eager_transforms.py
@@ -1305,6 +1305,19 @@ def _register_jit_decomposition(decomp, use_python=False):
     torch.jit._register_decomposition(decomp, graph)
 
 
+# use an alternate way to register an operator into the decomposition table
+# _register_jit_decomposition doesn't work for some operators, e.g. addr,
+#  because the Tensor types generated cannot be unioned by torchscript
+_register_jit_decomposition_lib = torch.library.Library("aten", "IMPL", "FuncTorchBatched")
+def _register_jit_decomposition_bypass_script(decomp):
+    if decomp in decomposition_table:
+        decomposition_table_used = decomposition_table
+        name =  decomp.overloadpacket.__name__
+        _register_jit_decomposition_lib.impl(name, torch._decomp.decomposition_table[decomp])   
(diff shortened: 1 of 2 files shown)


functorch/_src/eager_transforms.py
```diff
@@ -1305,6 +1305,20 @@ def get_function_def(sig):
...
 
 
+# use an alternate way to register an operator into the decomposition table
+# _register_jit_decomposition doesn't work for some operators, e.g. addr,
+#  because the Tensor types generated cannot be unioned by torchscript
+# decomp should be type OpOverload
+_register_jit_decomposition_lib = torch.library.Library("aten", "IMPL", "FuncTorchBatched")
+
+
+def _register_jit_decomposition_bypass_script(decomp):
```
Chillee, line 1315: call this `_register_python_decomposition_vmap` I think.

functorch/_src/eager_transforms.py
```diff
@@ -1305,6 +1305,20 @@ def get_function_def(sig):
     torch.jit._register_decomposition(decomp, graph)
 
 
+# use an alternate way to register an operator into the decomposition table
+# _register_jit_decomposition doesn't work for some operators, e.g. addr,
+#  because the Tensor types generated cannot be unioned by torchscript
+# decomp should be type OpOverload
+_register_jit_decomposition_lib = torch.library.Library("aten", "IMPL", "FuncTorchBatched")
```
Chillee, line 1312: This isn't using the `_jit` path - I'd just call it something like `vmap_decompositions_lib`.

Chillee (approved):
LGTM - just need to figure out the MacOS failure.

PR #844, opened 2022-05-31
Fix CI

zou3519:


functorch/csrc/DynamicLayer.cpp
```diff
@@ -479,8 +479,6 @@ TORCH_LIBRARY_IMPL(aten, FT_DYNAMIC_LAYER_FRONT_MODE_KEY, m) {
   JVP_DECOMP(_softmax_backward_data);
   OP_DECOMPOSE(log_sigmoid);
   JVP_DECOMP(log_sigmoid_forward);
-  JVP_DECOMP(binary_cross_entropy);
-  JVP_DECOMP(binary_cross_entropy_backward);
```
samdow, line 483: (I always forget about this one) we also have the list in eager_transforms that we should probably also update. Idk a great way to make sure we've actually gotten all the lists but I guess hopefully this is just a temporary hack and we can clean them all up once we're done

zou3519: Does anything else use https://github.com/pytorch/functorch/blob/28d3ef7f8f7819b630cd067cace95473eb520dad/functorch/_src/eager_transforms.py#L1330-L1331 ?

samdow: It shouldn't since I added it for the forward over reverse support https://github.com/pytorch/functorch/pull/791/files and it looks like it was added for the jvp decompositions (https://github.com/pytorch/functorch/commit/a1f569ac19f65bf5c5b1b6928f49f02a69204fcf)

But I'm not absolutely certain about this (cc @eellison @Chillee)

zou3519: From code reading I agree with your assessment @samdow, and fwiw tests pass, so I am merging this.

Issue #622, opened 2022-03-26
Does vmap not work on GroupNorm?


QiyaoWei:
Say I want to calculate the jacobian for one sample over ten output classification logits like this
```
import torch
import torch.nn as nn
import torch.nn.functional as F
from functorch import vmap

class Layer(nn.Module):
    def __init__(self, n_channels, n_inner_channels, kernel_size=3, num_groups=1):
        super().__init__()
        self.conv1 = nn.Conv2d(n_channels, n_inner_channels, kernel_size, padding=kernel_size // 2, bias=False)
        self.conv2 = nn.Conv2d(n_inner_channels, n_channels, kernel_size, padding=kernel_size // 2, bias=False)
        self.norm1 = nn.BatchNorm2d(n_inner_channels)
        self.norm2 = nn.BatchNorm2d(n_channels)
        self.norm3 = nn.BatchNorm2d(n_channels)
        self.conv1.weight.data.normal_(0, 0.01)
        self.conv2.weight.data.normal_(0, 0.01)
        self.linear = nn.Linear(3*32*32, 10)

    def forward(self, z, x=None):
        if x == None:
            x = torch.zeros_like(z)
        y = self.norm1(F.relu(self.conv1(z)))
        out = self.norm3(F.relu(z + self.norm2(x + self.conv2(y))))
        return self.linear(torch.flatten(out, 1))

x = torch.randn(5, 3, 32, 32)
model = Layer(3,3)

def jacobian(model, x_i, vs):
    output = model(x_i)
    def vjp(v):
        return torch.autograd.grad(output, model.parameters(), v.view_as(output))
    result = vmap(vjp)(vs)
    return torch.cat([g.flatten(1) for g in result], 1).detach()

vs = torch.eye(10)
# This is vmap
model.eval()
temp = jacobian(model, x[:1], vs)

#This is testing vmap output
output = model(x[:1])
jacobian_rows = [torch.autograd.grad(output, model.parameters(), v.view_as(model(x[:1])), retain_graph=True) for v in vs.unbind()]
jacobian = []
for jac in jacobian_rows:
    jacobian.append(torch.cat([g.flatten(0) for g in jac], 0).detach())
jacobian = torch.stack(jacobian)
assert torch.allclose(temp, jacobian)
```

The assertion works as expected. However, I change nothing else other than the norm function
```
import torch
import torch.nn as nn
import torch.nn.functional as F
from functorch import vmap

class Layer(nn.Module):
[...]

PR #851, opened 2022-06-02
Add exhaustive testing of vmap autograd composability

### LINKED ISSUE #704: Exhaustively check that vmap composes with autograd.grad
#630 adds scaffolding to test `vmap(f)` where `f` calls `autograd.grad` because an issue came up with group_norm when composing `vmap` and `autograd.grad`. We should exhaustively check that all functions with OpInfos work with this composition, building off of the scaffolding that exists

There's a dirty hack in the current group norm test that should go: batched tensors can't be passed as inputs to autograd.grad because they can't `requires_grad` so if we have that we just return a bunch of 1s or 0s (I forget which). Instead what we should be doing is taking the output from `get_autograd_fn_and_args_with_cotangents` and use a wrapper function so the only thing that can be batched are the cotangents (the first n arguments where n is the number of cotangents)

The failures of this should match exactly test_vmapvjp but there may be extra ones if we aren't handling backward formulas correctly but are handling the infinitely_differentiable versions right

samdow:
Fixes #704

Adds exhaustive testing of vmap(f) where f calls `torch.autograd.grad`. This caused some issues (#622) because functorch will set enable GradMode

We only allow batching the grad_output argument since everything else will require gradients


test/test_ops.py
```diff
@@ -199,28 +199,31 @@ def wrapped(*args):
...
     flat_cotangents, cotangents_spec = tree_flatten(cotangents)
 
     @functools.wraps(f)
-    def wrapped(*args):
-        assert len(args) == len(flat_args) + len(flat_cotangents)
-        actual_args = args[:len(flat_args)]
-        cotangents = args[len(flat_args):]
-        actual_args = tree_unflatten(actual_args, args_spec)
-        cotangents = tree_unflatten(cotangents, cotangents_spec)
+    def wrapped(*cotangents):
```
zou3519, line 206: The comment above `get_autograd_fn_and_args_with_cotangents` needs to be updated, right? This returns a new function that accepts cotangents and returns the vjp

zou3519: I'd also change the name of this function to reflect that it's returning something that only accepts cotangents

test/test_ops.py
```diff
@@ -1424,21 +1427,49 @@ def fn(input, weight, bias):
...
+    def test_vmap_autograd_grad(self, device, dtype, op):
+        if not op.supports_autograd:
+            self.skipTest("Skipped! Autograd not supported.")
+            return
 
+        sample_inputs = op.sample_inputs(device, dtype, requires_grad=True)
         for sample_input in sample_inputs:
             cotangents = get_sample_cotangents(op, sample_input)
             f, args = get_autograd_fn_and_args_with_cotangents(op, sample_input, cotangents)
             for loop_out, batched_out in get_fallback_and_vmap_exhaustive(f, args, {}, opinfo=op):
```
zou3519, line 1472: Actually, I'm a little confused. Why isn't this just:
```
output = op(*sample_input.args, **sample_input.kwargs)
cotangents = tree_map(torch.randn_like, output)
def compute_grad(cotangents):
  return torch.autograd.grad(get_differentiable(output), get_differentiable(sample_inputs), cotangents, retain_grad=True)
 
for loop_out, batched_out in get_fallback_and_vmap_exhaustive(compute_grad, cotangents, {}, opinfo=op):
```
This gives us a couple of benefits:
1. avoids re-computing the forward pass multiple times
2. I'm actually not sure what get_autograd_fn_and_args_with_cotangents is doing, but is this similar?

samdow:
Ah I see. That makes a lot of sense and actually saved a lot of xfails since most of the functions with randomness errors (dropout, bernoulli, etc) were coming from the forward pass. Now that we aren't vmapping over the forward invocation, they work. Thanks!

test/test_ops.py
```diff
@@ -1433,21 +1404,68 @@ def fn(input, weight, bias):
...
+                    out_and_cotangents = [(o, cotangent) for (o, cotangent) in out_and_cotangents if is_differentiable(o)]
+                    outs, cotangents = zip(*out_and_cotangents)
+
+                return filter_none(
+                    torch.autograd.grad(outs, get_flat_differentiable(primals), cotangents,
+                                        retain_graph=True, allow_unused=True))
+
+            for loop_out, batched_out in get_fallback_and_vmap_exhaustive(compute_grad, (cotangents,), {}, opinfo=op):
+                self.assertEqual(loop_out, batched_out)
                 self.assertEqual(loop_out, batched_out)
```
zou3519, line 1469: self.assertEqual happens twice?

test/test_ops.py
```diff
@@ -1433,21 +1404,68 @@ def fn(input, weight, bias):
...
                 self._compare_jacobians_of_vjp(fn, (cotangents, input, weight, bias))
 
-    @ops(filter(lambda op: op.name == "nn.functional.group_norm", functorch_lagging_op_db + additional_op_db),
-         allowed_dtypes=(torch.float32, torch.double))  # TODO: generalize
-    def test_group_norm_backward(self, device, dtype, op):
-        # hacky, only works since no group norm inputs can be scalars
-        def was_skipped_from_batched_tensors(batched_out, batch_size):
-            return batched_out.shape == (batch_size,) and all(tuple(e == 1 for e in batched_out))
+    @skipOps('TestOperators', 'test_vmap_autograd_grad', {
+        # randomness errors
```
zou3519, line 1408: nit: nothing in this section anymore, should we delete this line?

zou3519 (approved):
LGTM, some minor comments. Thanks for doing this, it looks like our coverage on this is decent

zou3519:
CUDA build failures have been fixed on main, you might want to rebase

PR #856, opened 2022-06-07
Separate forward and backwad compilation for default partition

anjali411:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* (to be filled)

### DIFF
--- a/functorch/_src/aot_autograd.py
+++ b/functorch/_src/aot_autograd.py
@@ -53,6 +53,7 @@ aten = torch.ops.aten
 
 
 def create_joint_forward_backward(fn):
+    # tangents are just grad_outs/cotangents (wrong naming)
     def joint_forward_backward(
         primals: List[Any], tangents: List[Any]
     ) -> Tuple[List[Any], List[Any]]:
@@ -140,12 +141,14 @@ def create_aot_autograd_function(
     compiled_fw = None
     compiled_bw = None
     num_outs = None
-
+    joint_inputs = None
+    fw_outs = None
+    aot_decompositions = {**aot_autograd_decompositions, **decompositions}
     class CompiledFunction(torch.autograd.Function):
         @staticmethod
         @disable_torchdynamo
         def forward(ctx, *flat_tensor_args):
-            nonlocal compiled_fw, compiled_bw, num_outs
+            nonlocal compiled_fw, num_outs, joint_inputs, fw_outs
             if compiled_fw is None:
                 with torch.set_grad_enabled(grad_state):
                     out = flat_fn(*flat_tensor_args)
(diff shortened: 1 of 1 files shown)


functorch/_src/aot_autograd.py
```diff
@@ -140,12 +140,15 @@ def create_aot_autograd_function(
     compiled_fw = None
     compiled_bw = None
     num_outs = None
+    joint_inputs = None
```
anjali411, line 143: need to save these tensors in the context

functorch/_src/aot_autograd.py
```diff
@@ -159,31 +160,83 @@ def forward(ctx, *flat_tensor_args):
...
+                    # assert that the forward graph generated here is the same
+                    # if it's specified that the user might want to calculate double backwards
+                fw_module, bw_module = partition_fn(fx_g, ctx.saved_tensors[ctx.num_intermediate:])
+                print(fw_module.code)
+                print(ctx.fwd_graph)
+                assert fw_module.code == ctx.fwd_graph
+                func_code = bw_module.code.split('self, ')
+                # print(func_code[0] + func_code[1])
+                exec(func_code[0] + func_code[1], globals())
+                f = create_aot_autograd_function(forward, bw_compiler, bw_compiler, partition_fn, aot_decompositions, grad_state)
```
zou3519, line 220: Two questions:
- Why are we passing `forward` to create_aot_autograd_function? I would have expected us to pass bw_module.code without the self argument
- What is the exec for? Are you trying to test this without the `create_aot_autograd_function` line?

anjali411: - forward is the name of the function generated by running `bw_module.code`
- exec executes the bw_module.code to create a backward function which is the forward for the next pass

functorch/_src/aot_autograd.py
```diff
@@ -196,33 +206,76 @@ def forward(ctx, *flat_tensor_args):
...
+                inp_grad_outs = flat_grad_outs
+                with torch.set_grad_enabled(grad_state):
+                    fx_g_b = make_fx(j_b, aot_decompositions)(flat_tensor_args, inp_grad_outs)
+                    if config.use_functionalize:
+                        # Functionalize the foward backward graph. First create a
+                        # fake fn to make functionalize happy
+                        def fake_fn(primals, tangents):
+                            return fx_g_b(primals, tangents)
+                        fx_g_b = make_fx(functionalize(fake_fn))(flat_tensor_args, inp_grad_outs)
+                saved_value_nodes = _get_saved_values(fx_g_b, saved_value_names)
```
anjali411, line 245: Unfortunately this approach doesn't always work because the newly generated fx graph may not have the same nodes as the previous graph. We need an alternate way to select nodes of interest in this new graph!

PR #860, opened 2022-06-07
Fix MSE foward, use decomposition for MSE backward

### LINKED ISSUE #858: jacrev(vjp(mse_loss)) raises an error
https://gist.github.com/zou3519/76770e6f9798a406028fda34773c5ff9

Found while looking through our CI failures. Why didn't this fail before?

samdow:
Because MSE backward was redispatching, it ended up having some shape errors if it redispatched twice. This switches it to use the decomposition from core so that we don't try to guess based on the size of the input if it came from a redispatch or an original call. Specifically:

- when reduction != none, the original output would be of size [N]. We would be padding the grad_output with 1s if need be and then redispatching (with reduction == none). This led to issues because reduction == none assumes that grad output and self have the same number of elements
- when reduction == none, often times the grad_output would be flattened (a little unclear why this was happening, tbh). This was causing reshape errors. If we tried to fix this by just reshaping grad_output to match self, we would end up with errors from the redispatch because grad_output wouldn't have enough elements in that case

There were also some issues in the forward rule where if reduction was none, the output would be flattened where it shouldn't be. Specifically, if you had
```
input: [B, in_0, in_1, ...]
target: [B, in_0, in_1, ...]
output (should be): [B, in_0, in_1, ...]
output (produced): [B, in_0 * in_1 * ...]
```

This was not caught because all op info tests that had a none reduction only had 1D inputs. This adds new tests and fixes the forward + backward formula

Fixes #858

### DIFF
--- a/functorch/csrc/BatchRulesLoss.cpp
+++ b/functorch/csrc/BatchRulesLoss.cpp
@@ -52,15 +52,28 @@ mse_loss_backward_batch_rule(
   auto grad_output_ = moveBatchDimToFront(grad_output, grad_output_bdim);
   auto self_ = moveBatchDimToFront(self, self_bdim);
   auto target_ = moveBatchDimToFront(target, target_bdim);
-  if (reduction != Reduction::None && grad_output_bdim.has_value()) {
-    // grad_output_ is of shape [N]. Input is of shape [N?, ...].
-    // We need to view grad_output_ as shape [N, ...].
-    auto self_rank_without_bdim = rankWithoutBatchDim(self, self_bdim);
-    DimVector view_shape(self_rank_without_bdim + 1, 1);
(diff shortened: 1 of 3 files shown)


Chillee:
Can we just directly use the decomposition instead of writing a batching rule for it? It doesn't look like the batching rules is doing any fusion of the ops, so I don't think the perf will be significantly different.

samdow:
> Can we just directly use the decomposition instead of writing a batching rule for it?

Sure. The issue right now is that the decomposition in core is wrong in the case where reduction=none because of some shaping stuff. It will end up with the same errors theseus was seeing. So I'll need to add a test for that case in core, fix it, and then set up the infra here to use that. Because we're coming up on the release, I'm thinking that I'll merge this so that we have a fix that will definitely be in the release and then updating the decomposition

Issue #859, opened 2022-06-07
getitem batching rule failure


zou3519:
repro:
```
import torch
import torch.autograd.forward_ad as fwAD
from functorch import make_fx, vmap

x = torch.arange(3 * 5 * 5 * 5).view(3, 5, 5, 5)

def select(x):
    return x[(0, 0), :, (1, 1)]

y = vmap(select)(x)
print(y.shape)

print(select(x[0]).shape)

y = make_fx(vmap(select))(x)
print(y.code)
```
gonna xfail for now

zou3519:
List of failing tests:
```
FAILED test/test_ops.py::TestOperatorsCPU::test_vmapjvp___getitem___cpu_float32 - RuntimeError: Trying to set a forward gradi..
FAILED test/test_ops.py::TestOperatorsCPU::test_vmapjvpall___getitem___cpu_float32 - AssertionError: The values for attribute..
FAILED test/test_ops.py::TestOperatorsCPU::test_vmapjvpall_has_batch_rule___getitem___cpu_float32 - RuntimeError: Trying to s..
FAILED test/test_ops.py::TestOperatorsCUDA::test_vmapjvp___getitem___cuda_float32 - RuntimeError: Trying to set a forward gra..
FAILED test/test_ops.py::TestOperatorsCUDA::test_vmapjvpall___getitem___cuda_float32 - AssertionError: The values for attribu..
FAILED test/test_ops.py::TestOperatorsCUDA::test_vmapjvpall_has_batch_rule___getitem___cuda_float32 - RuntimeError: Trying to..
FAILED test/test_vmap.py::TestVmapOperatorsOpInfoCPU::test_vmap_exhaustive___getitem___cpu_float32 - AssertionError: The valu..
FAILED test/test_vmap.py::TestVmapOperatorsOpInfoCUDA::test_vmap_exhaustive___getitem___cuda_float32 - AssertionError: The va..
```

PR #862, opened 2022-06-08
Fix index.Tensor, index_put batching rules

zou3519:
Fixes https://github.com/pytorch/functorch/issues/859

Start reading at `NOTE: [advanced indexing (index.Tensor) batch rule]`
in the code for details. This PR rewrites the index.Tensor and index_put
batching rules.

The TL;DR is:
- advanced indexing has different behavior depending on if the "advanced
indices are adjacent":
https://numpy.org/doc/stable/user/basics.indexing.html#combining-advanced-and-basic-indexing
- we have to take this into account in our batching rules, because
index.Tensor and index_put handle these internally.

Test Plan
- I added new test cases for getitem and aten.ops.index_put via OpInfo
testing.

Future
- primtorch should have a sane decomposition that we can use
- We haven't fixed the index_put_ batching rule yet. TODO later...
- Upstream our test cases (see next section) into pytorch/pytorch

### DIFF
--- a/functorch/csrc/BatchRulesScatterOps.cpp
+++ b/functorch/csrc/BatchRulesScatterOps.cpp
@@ -16,7 +16,45 @@
 
 namespace at { namespace functorch {
 
-std::tuple<std::vector<optional<Tensor>>, int64_t, int64_t> batchIndices(
+static bool any_has_value(ArrayRef<optional<int64_t>> bdims) {
+  for (const auto& bdim : bdims) {
+    if (bdim.has_value()) {
+      return true;
+    }
+  }
+  return false;
+}
+
+static int64_t get_num_leading_nones(ArrayRef<optional<Tensor>> indices) {
+  int64_t result = 0;
+  for (const auto& idx : indices) {
+    if (!idx.has_value() || !idx->defined()) {
+      result++;
+    } else {
+      return result;
+    }
+  }
+  return result;
+}
+
+static int64_t get_max_index_logical_dim(
+    ArrayRef<optional<Tensor>> indices,
+    ArrayRef<optional<int64_t>> indices_bdims) {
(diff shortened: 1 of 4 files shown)


functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -102,25 +155,152 @@ std::tuple<Tensor,optional<int64_t>> index_batch_rule(
     ArrayRef<optional<Tensor>> indices,
     ArrayRef<optional<int64_t>> indices_bdims) {
 
+  // NOTE: [advanced indexing (index.Tensor) batch rule]
```
zou3519, line 158: Start reading here

functorch/csrc/BatchRulesScatterOps.cpp
```diff
@@ -102,25 +155,152 @@ std::tuple<Tensor,optional<int64_t>> index_batch_rule(
...
+  std::iota(permutation.begin(), permutation.end(), 0);
+  std::iota(
+      permutation.begin() + 1,
+      permutation.begin() + 1 + num_leading_nones,
+      1 + max_index_dim);
+  std::iota(
+      permutation.begin() + 1 + num_leading_nones,
+      permutation.begin() + 1 + max_index_dim + num_leading_nones,
+      1);
+  return std::make_tuple(res.permute(permutation), 0);
```
samdow, line 303: nit: since we're doing the same thing for this here and in maybe_permute values and it has such a thorough explanation, it might be nice to have this as a helper function

samdow (approved):
What an adventure in batched indices. This is truly wild

functorch/csrc/BatchRulesScatterOps.cpp
zou3519: Good idea

PR #863, opened 2022-06-08
Fix CI by using pip instead of conda to install torch

zou3519:


zou3519:
Seems to be working: 
image

PR #864, opened 2022-06-09
Fix cuda tests, skip mac unexpected successes

samdow:
Moving back to conda after it failed to build for a day. Seems like a transient problem

### DIFF
--- a/.circleci/unittest/linux/scripts/install.sh
+++ b/.circleci/unittest/linux/scripts/install.sh
@@ -59,5 +59,6 @@ if [ "${CU_VERSION:-}" == cpu ] ; then
     PYTORCH_VERSION="$(python -c "import torch; print(torch.__version__)")" python setup.py develop bdist_wheel -d $WHEELS_FOLDER
 else
     pip install torch torchvision -f https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html --pre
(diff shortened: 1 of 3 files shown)


.circleci/unittest/linux/scripts/install.sh
```diff
@@ -59,5 +59,6 @@ if [ "${CU_VERSION:-}" == cpu ] ; then
     PYTORCH_VERSION="$(python -c "import torch; print(torch.__version__)")" python setup.py develop bdist_wheel -d $WHEELS_FOLDER
 else
     pip install torch torchvision -f https://download.pytorch.org/whl/nightly/cu102/torch_nightly.html --pre
+    conda install pytorch cudatoolkit=10.2 -c pytorch-nightly
```
zou3519, line 62: Wait, does this completely override the `pip install torch` in the previous line?

If so, then we could just delete the `pip install torch`. It used to actually be conda, but that ran into a weird failure a couple of days ago

samdow: Let me try without pip and see what's happening. I'm guessing that it actually doesn't override pip because when I was playing around with it on the SSH version, I actually saw that after doing this, the conda env would still point to the pip-ed version of pytorch. However, it does feel really hacky this way--fingers crossed that the conda error was just some transient thing

PR #868, opened 2022-06-11
Utilities to get GPU Utilization from chrome trace dumps

yushangdi:


yushangdi:
A commandline tool that processes the chrome traces output by the pytorch profiler, and outputs the GPU Utilization and the percentage of program time spent in GPU and matmal/conv operations.

yushangdi:
Example usage:

`python chrome_trace_parser.py -f [filename] -runtime [runtime-file]`

The filename should be in format [modelname]_chrome_trace*.json
It will process the chrome trace in [filename]. The [runtime-file] is a csv file that contains the program total time, in the format of 

name, runtime
[modelname], [time in second]

modelname should match the name of the chrome trace file. 


One can also use 

`python chrome_trace_parser.py -fd [folder] -runtime [runtime-file]`

to get the results for all json files in folder [folder]

benchmarks/chrome_trace_parser.py
```diff
@@ -0,0 +1,121 @@
...
+            mm_conv_utilization = get_duration(sorted_gpu_mm_conv_events) / total_length
+
+            print(f"{modelname}, {utilization}, {mm_conv_utilization}")
+        except:
+            logging.exception(f"{filename}, ERROR")
+            print(f"{filename}, ERROR")
+
+
+if __name__ == "__main__":
+    main()
```
Chillee, line 121: should have a newline here.

Chillee (approved):
LGTM - minor nit about formatting, but good to merge.

I think we should end up wrapping these couple of PRs together, but that can be left for future work.

benchmarks/chrome_trace_parser.py
yushangdi: added

PR #875, opened 2022-06-14
Removed some obsolete workarounds in ts_compile and added a new one

Chillee:


functorch/_src/python_key.py
```diff
@@ -7,3 +7,4 @@
 from torch.fx.experimental.proxy_tensor import make_fx, ProxyTensor, dispatch_trace, PythonKeyTracer, decompose
 
 pythonkey_decompose = decompose
+PythonTensor = ProxyTensor
```
jjsjann123, line 10: Is this change intended?

ngimel: Yeah, that's to unbreak dynamo

PR #675, opened 2022-04-07
Add decomposition for `aten.native_batch_norm_backward` op

Shukla-Gaurav:
This commit adds decomposition for the `aten.native_batch_norm_backward`
op.

Signed-Off-by: Gaurav Shukla <gaurav@nod-labs.com>

PR #877, opened 2022-06-15
[jvpvjp] Batch norm coverage with decomposition

samdow:
Adds decomposition from #675 for forward over reverse coverage. Similar to with layer norm, we needed to recompute the mean and variance so autograd propagates properly (sad) and needed to return tensors of zeros instead of None (sad)


zou3519 (approved):

functorch/_src/decompositions.py
```diff
@@ -128,3 +134,84 @@ def native_layer_norm_backward(
...
+        grad_weight = dot_p * invstd
+    elif weight is not None:
+        grad_weight = torch.zeros_like(weight)  # should be None but doesn't work with vjp
+    else:
+        grad_weight = torch.zeros(())  # should be None but doesn't work with vjp
+
+    if output_mask[2]:
+        grad_bias = grad_output_sum
+    else:
+        grad_bias = torch.zeros_like(grad_output_sum)  # should be None but doesn't work with vjp
```
zou3519, line 215: sad, but is it what it is

PR #878, opened 2022-06-15
Remove pad_circular jvp x vjp xfail

zou3519:
For reasons unknown to me, pad_circular ends up mutating the tangent on
the grad output. Perhaps this is a bug (see minimum repro over at
https://gist.github.com/zou3519/9d2a4e49faf63b3495733df9777b95f8): pad
is an out-of-place operation, so on gradient computation, it should
really not be modifying the gradient's tangent in-place?

The fix is simple: we just need to clone the grad_output's tangents
before passing it to our tests that compare functorch's jvp(vjp against
pytorch's forward-over-reverse support.


zou3519:
This should instead be fixed by https://github.com/pytorch/pytorch/pull/79653

zou3519:
Not needed anymore

PR #899, opened 2022-06-22
add a pass in ts_compile to prepare for jit.script

yushangdi:


functorch/_src/compilers.py
```diff
@@ -20,6 +20,12 @@ def _canonicalize(fx_g):
     return fx_g
 
 
+def _prepare_for_jit_script(gm):
```
Chillee, line 23: Can we rename this to something like `strip_overloads`, add a docstring, and also add a `gm.recompile()`?

yushangdi: added

PR #917, opened 2022-06-28
Use circleci to build linux binaries

zou3519:


zou3519:
Tried this out, it didn't seem to work

PR #919, opened 2022-06-28
Attempt build without std::cout

zou3519:


zou3519:
Didn't work. Probably because functorch still uses ostream

PR #935, opened 2022-07-06
Use fake tensor for primal computation in AOTAutograd

ezyang:
This prevents AOTAutograd from mutating inputs multiple times
when the internal function mutates its inputs.


ezyang:
Blocked on a nightly with https://github.com/pytorch/pytorch/pull/80995

Chillee:
@ezyang I was running into some issues turning on FakeTensor for AOTAutograd by default. It ended up throwing a bunch more errors on torchbench with torchdynamo.

I'll dig into those today.

ezyang:
I have a cluster of bug fixes which should help. I'll add you on them. They're consolidated in the cuda graphs backend poc

ezyang:
I guess I can't land this without the bugfixes in https://github.com/pytorch/pytorch/pull/81192

byronyi:
So I guess issues in #802 will also be fixed with this PR.

ezyang:
hmm I guess the fix missed the most recently nightly

ezyang:
To move this along, I've disabled fake tensor mode globally for now, it will be easy to flip it on for testing

Issue #867, opened 2022-06-11
Why is using vmap(jacrev) for BatchNorm2d in non-tracking mode not working?


kwmaeng91:
Hi, experts.
I am trying to use vmap(jacrev) to calculate the per-sample jacobian in a batch for my network during inference. However, when there is BatchNorm2d, it does not work. Because during inference, BatchNorm2d is simply applying the statistics previously tracked (and not doing any inter-sample operations), I think it should work just as any other simple operation from my understanding. Is there a way for me to make it work, or is there anything I am misunderstanding?

Below is my minimal code:
```
from functorch import jacrev, vmap
import torch
from torch import nn

layers = nn.Sequential(
        nn.Conv2d(3, 3, kernel_size=(3, 3)),
        nn.BatchNorm2d(3, track_running_stats=False),
    )

x = torch.randn(4, 3, 30, 30)
j = vmap(jacrev(layers))(x)
```

And I get this error in the bn layer
`ValueError: expected 4D input (got 3D input)`

I think this should fundamentally be doable, and just might be because of how vmap and jacrev is implemented.
Is there any simple workaround, or am I misunderstanding anything?

Thank you for any help

kwmaeng91:
Dear @samdow, Thank you for your reply! It is really helpful.
I have a followup question. Now it works with `track_running_stats=False` with your help, but it still does not with `track_running_stats=True`, even when the model is in eval mode (i.e., `net.eval()`). I thought something similar will work as well for eval(), but I guess it does not?

From here (https://pytorch.org/functorch/stable/batch_norm.html), it says "Batch Norm requires in-place updates to running_mean and running_var of the same size as the input. Functorch does not support inplace update to a regular tensor that takes in a batched tensor (i.e. regular.add_(batched) is not allowed)." as the reason why this does not work. However, doesn't running_mean and running_var **not get updated in eval() mode**?? I am not sure why vmap cannot work with eval() mode.

Will there be yet another fix for `track_running_stats=True` with eval mode (i.e., `net.eval()`, or is this fundamentally not doable right now unlike my original question?
I can make this a separate issue if that makes things more clear.

Thank you for the help!

samdow:
Hi @kwmaeng91! Great catch! I believe that this is a bug that we should fix. I may not have bandwidth to do this for a bit, but given your use case, I'll try to prioritize it. The fix itself should be simple but we'll want to test it and the way we're testing batch norm is a little subtle to avoid the error case

[...]

PR #951, opened 2022-07-11
Generate n^2 not n^3 inputs for batch and instance norm; small batch norm fix

samdow:
Follows the same idea as #937 but for batch norm and instance norm and does a little refactor to get rid of a TODO

In this something weird happened and it looks like we weren't testing fully for `vmapjvpall` with batch norm before...basically it's now failing because it's trying to differentiate wrt running_mean and running_var

Also this surfaced a typo in the batch norm batching rule and an issue where we were expanding running_mean or running_var without making them contiguous, caused errors when cuda tried to write to that tensor. Both of these errors were fixed in here to keep CI green

### DIFF
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -70,6 +70,34 @@ def get_bdim_choices(num_tensors):
     assert choices[-1] == (None,) * num_tensors
     return tuple(choices[:-1])
 
+# NB: This is O(2 ** num_tensors).
+# num_tensors ranges from 1 to 10, with 2-4 being most common.
+# Try not to extravagate it if you're modifying it.
+def get_bdim_choices_batch_norm(num_tensors, _, running_mean=None, running_var=None, *args):
+    choices = []
+    options = (-1, None)
+
+    # instance norm turns these into unbatched 0 tensors, so we cannot batch the input if either is not specified
+    if running_mean == None or running_var == None:
+        choices.append((None,) + (0,) * (num_tensors - 1))
+        for choice in itertools.product(options, repeat=num_tensors - 1):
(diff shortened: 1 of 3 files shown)


test/common_utils.py
```diff
@@ -120,87 +143,41 @@ def get_batched_arg(arg, bdim):
...
-
-            if all([i is None for i in in_dims]):
-                continue
-
-            yield pytree.tree_unflatten(batched_args, arg_spec), pytree.tree_unflatten(in_dims, arg_spec), kwarg_values
+def get_exhaustive_batched_inputs_batch_norm(arg_values, kwarg_values, batch_size=2):
+    flat_args, arg_spec = pytree.tree_flatten(tuple(arg_values))
+    is_tensors = [isinstance(a, torch.Tensor) for a in flat_args]
+    num_tensors = sum(is_tensors)
+    if num_tensors == 1:  # if there's only an input, we can't test :/
```
zou3519, line 150: Why not?

samdow: Ahh I added a better comment and then never pushed that change. This will give a little more color though

Basically if there's only an input (running_mean and running_var are None) and we're training, the function will turn the Nones into tensors filled with 0s. Because these won't be batched tensors, if we batch the input we'll fail. Since there's just an input, the only interesting case is the one where the input is batched which we can't test

After this PR, I want to add one that lets us test all variants of batching if `training=False` per #867 so there we would test what happens if we batch with just an input

zou3519 (approved):

test/common_utils.py
zou3519: Good point, thanks for the explanation

PR #965, opened 2022-07-14
add view.SymInt kernel

Krovatkin:
This PR introduces `view.SymInt` overload to functorch ahead of time, so https://github.com/pytorch/pytorch/pull/81145 can pick it up and hopefully to not break functorch tests.

### DIFF
--- a/functorch/csrc/BatchRulesViews.cpp
+++ b/functorch/csrc/BatchRulesViews.cpp
@@ -11,6 +11,7 @@
 #include <functorch/csrc/PlumbingHelper.h>
 #include <functorch/csrc/BatchedFallback.h>
 #include <ATen/core/dispatch/Dispatcher.h>
+#include <ATen/core/TensorBody.h>
 #include <c10/core/SymIntArrayRef.h>
 #include <c10/util/SmallBuffer.h>
 #include <ATen/InferSize.h>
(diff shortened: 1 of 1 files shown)


functorch/csrc/BatchRulesViews.cpp
```diff
@@ -435,14 +486,28 @@ std::tuple<Tensor, optional<int64_t>> view_batching_rule(
   return std::make_tuple(self_.view(size_), 0);
 }
 
+std::tuple<Tensor, optional<int64_t>> view_symint_batching_rule(
+    const Tensor& self,
+    optional<int64_t> self_bdim,
+    c10::SymIntArrayRef size) {
+  return self.view(self_bdim, c10::asIntArrayRefSlow(size));
+}
```
zou3519, line 494: You do want to do what is done in [this gist](https://gist.github.com/zou3519/75b446ec3627a783cebe76fc61aa2e54) because m.impl will require that the function passed to it (view_symint_batching_rule) has the same schema as the operator (view.SymInt) when it exists.

The VMAP_SUPPORT macro does some template magic that makes it so that the function passed has a different signature than the Schema

Krovatkin: Got it! Copy-pasted yr version!

functorch/csrc/BatchRulesViews.cpp
```diff
@@ -435,6 +437,18 @@ std::tuple<Tensor, optional<int64_t>> view_batching_rule(
...
+Tensor view_symint_decomposition(const Tensor& self,
+            c10::SymIntArrayRef size) {
+  return at::view(self, c10::asIntArrayRefSlow(size));
+}
+
+std::tuple<Tensor, optional<int64_t>> view_symint_batching_rule(
+    const Tensor &self, optional<int64_t> self_bdim, c10::SymIntArrayRef size)
+{
+  return self.view(self_bdim, c10::asIntArrayRefSlow(size));
+}
```
zou3519, line 449: This isn't necessary anymore (and is causing the build failures)

zou3519 (approved):
After we delete the unnecessary function we should be good to go. I would wait for the test suite to run a little just in case the dispatcher having something registered to a non-existing operator gives us test-time problems

PR #978, opened 2022-07-21
Remove eig mentions in test

IvanYashchuk:
This PR is needed because tests are failing when trying to remove deprecated `torch.eig`, see https://github.com/pytorch/pytorch/pull/70982#issuecomment-1191783202.


IvanYashchuk:
@zou3519, what should be the order of these PRs? Since it's removing xfails I assume tests start to fail, but https://github.com/pytorch/pytorch/pull/70982's functorch CI is red without this change.

zou3519:
@IvanYashchuk please hold off on merging this PR for one day. We're about to merge functorch into the pytorch/pytorch repository; when that happens, then you can just modify the functorch inside the pytorch/pytorch repository.

IvanYashchuk:
Sounds good!

Issue #273, opened 2021-11-16
Test functorch transform behavior on BatchNorm and other weird in-place operations


zou3519:
Most in-place operations look the same: the first argument is mutable and no other arguments are mutable.

F.batch_norm is an exception to that rule: it has multiple mutable arguments. We should check that our transforms work correctly on it.

ain-soph:
Hi @zou3519 , is this error related to the current issue? It seems all models with batchnorm can't be used for jacobian calculation.

```python3
import torch
import torch.nn as nn
import functorch
from torch.nn.utils import _stateless
import functools

model = nn.Sequential(nn.Linear(3, 4), nn.BatchNorm1d(4), nn.Linear(4, 5), nn.Softmax(dim=-1))
_input = torch.randn(32, 3)
weight = model[0].weight.clone()
weight.requires_grad_()


def func(param: torch.Tensor, _input: torch.Tensor = None):
    _output: torch.Tensor = _stateless.functional_call(
        model, {'0.weight': param}, _input)
    return _output  # (N, C)


jac: torch.Tensor = functorch.jacrev(functools.partial(func, _input=_input))(weight)
```

```
Traceback (most recent call last):
  File "C:\Users\ain-s\OneDrive\workspace\trojanzoo\test.py", line 33, in <module>
    jac: torch.Tensor = functorch.jacrev(functools.partial(func, _input=_input))(weight)
  File "C:\Users\ain-s\miniconda3\lib\site-packages\functorch\_src\eager_transforms.py", line 441, in wrapper_fn
    vjp_out = vjp(f_wrapper, *primals, has_aux=has_aux)
  File "C:\Users\ain-s\miniconda3\lib\site-packages\functorch\_src\eager_transforms.py", line 270, in vjp
    primals_out = func(*diff_primals)
  File "C:\Users\ain-s\miniconda3\lib\site-packages\functorch\_src\eager_transforms.py", line 624, in f_wrapper
    return f(*replaced_args)
  File "C:\Users\ain-s\OneDrive\workspace\trojanzoo\test.py", line 28, in func
    _output: torch.Tensor = _stateless.functional_call(
  File "C:\Users\ain-s\miniconda3\lib\site-packages\torch\nn\utils\stateless.py", line 141, in functional_call
    out = module(args, **kwargs)
  File "C:\Users\ain-s\miniconda3\lib\site-packages\torch\nn\modules\module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
[...]

PR #994, opened 2022-08-04
0.2.1 batch norm fixes

samdow:
Adds two commits for batch norm bug fixes. From discussion on #273 and #867

cc @zou3519 I ended up cherry-picking two commits since #958 also caught a typo and need for a contiguous call. However, lines-wise, this ends up being a lot of testing changes. I think I can refactor both changes to have fewer testing changes but will end up looking different than the version on main. Any thoughts as to if I should keep it like this or try to refactor for fewer test changes?

### DIFF
--- a/functorch/csrc/BatchRulesNorm.cpp
+++ b/functorch/csrc/BatchRulesNorm.cpp
@@ -58,7 +58,7 @@ batch_norm_batch_rule(
   auto running_mean = *running_mean_maybe_owned;
   c10::MaybeOwned<Tensor> running_var_maybe_owned = at::borrow_from_optional_tensor(running_var_opt);
   auto running_var = *running_var_maybe_owned;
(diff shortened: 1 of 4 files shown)


samdow:
Change: since it's failing so many tests I'm going to go with option 1 and try to re-cherry-pick with very minimal testing changes

zou3519:
> Change: since it's failing so many tests I'm going to go with option 1 and try to re-cherry-pick with very minimal testing changes

Sounds good

zou3519:
I wouldn't worry about the mac signal -- the macos signal is failing because they deprecated the xcode image I think

zou3519 (approved):
code lgtm, let's wait for tests to pass

samdow:
Circle CI shows passing for [all jobs but mac](https://app.circleci.com/pipelines/github/pytorch/functorch/3649/workflows/b26a7c68-58fc-4bde-8371-f89c1a4da48d). Looks like github checks aren't updating

PR #1008, opened 2022-08-25
Create dummy package for functorch

zou3519:
Soon, `import functorch` will work after one installs PyTorch. However,
we want to maintain backwards compatibility for installing functorch
from source (because many folks do that).

This PR:
- deletes most of the code in functorch (which has already been
moved to the pytorch/pytorch repo)
- sets up a "dummy package" -- installing functorch from source will
install something that has no imports. `import functorch` will work by
virtue of the functorch dummy package requiring torch as a requirement,
and the torch binary will enable `import functorch`.
- leaves the README.md and some other necessary metadata (license,
CONTRIBUTING.md, etc) intact.

Test Plan:
- I tested that one is able to build a dummy package and upload to
testpypi
- We should test this in conjunction with
https://github.com/pytorch/pytorch/pull/83464


zou3519:
NB: the merge order looks something like:
- turn off shipit sync to pytorch/functorch
- merge stack at https://github.com/pytorch/pytorch/pull/83464
- when https://github.com/pytorch/pytorch/pull/83464 hits the pytorch nightly binary, then merge this

zou3519:
https://github.com/pytorch/functorch/pull/1017 is the updated version

